PR #1, opened 2019-05-11
Initial set-up

mostaphaRoudsari:
@chriswmackey I still need to:

 - [x] finish `face_to_idf` method.
- [x] add one or two tests.
<s> - [ ] add automated documentation </s>

PS: Changed my mind! I will send a separate PR for <s>tests and</s> documentation.


chriswmackey:
@mostaphaRoudsari . Thanks for putting this together and I will write more detailed comments soon.  
The main question that I have is why we are writing a translator to go from the face to an IDF?  Aren't we planning to just use OpenStudio in [+]?

Also, you will see that I added a [Polyface object](https://github.com/chriswmackey/ladybug-geometry/blob/development/ladybug_geometry/geometry3d/polyface.py) to the geometry library that can handle all of the requirements and checks needed of honeybee zones (including checking whether the zone is closed and ensuring that all faces of the polyface are facing outward of the zone).  I mention this because maybe the translator should be from a polyface to an idf zone?

mostaphaRoudsari:
> The main question that I have is why we are writing a translator to go from the face to an IDF? Aren't we planning to just use OpenStudio in [+]?

I used the translator as an excuse to show how one can extend the `to` method in `honeybee.face.Face`. It will eventually be replaced by OpenStudio code assuming we can find a solution to only translate a single Poly/Face.

> Also, you will see that I added a Polyface object to the geometry library that can handle all of the requirements and checks needed of honeybee zones (including checking whether the zone is closed and ensuring that all faces of the polyface are facing outward of the zone).

Thank you for the note. I will keep that in mind.

honeybee_energy/properties.py
```diff
@@ -0,0 +1,46 @@
+"""Energy Properties."""
+from .boundarycondition import BoundaryConditions
+
+
+class EnergyProperties(object):
```
chriswmackey: Am I correct in understanding that this gets assigned to the Honeybee Face as a property?

honeybee_energy/properties.py
```diff
@@ -0,0 +1,46 @@
...
+    """Energy Properties for Honeybee geometry."""
+
+    def __init__(self, srf_type):
+        self.construction = srf_type.energy_construction
+        self.boundary_condition = BoundaryConditions().outdoors
```
chriswmackey: Do wee need to specify a separate construction and boundary condition instance for every single face?  This seems like it will greatly increase the memory needed when you have thousands of surfaces with the same construction and boundary condition. Is there some way that we could just create one instance for each of the default boundary conditions and constructions and have the various faces reference that instance?

honeybee_energy/properties.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @property
+    def construction(self):
+        """Set or get Face construction."""
+        return self._construction
+
+    @construction.setter
+    def construction(self, value):
+        self._construction = value
```
chriswmackey: Don't we need an `assert isinstance(value, Construction)` here?

honeybee_energy/properties.py
```diff
@@ -0,0 +1,46 @@
...
+    
+    @property
+    def materials(self):
+        """List of materials."""
+        if not self._construction:
```
chriswmackey: Is it possible to initialize energy properties without a construction?

honeybee_energy/faceutil.py
```diff
@@ -0,0 +1,32 @@
...
+            self.properties.face_type.name,
+            ep.construction.name,
+            self.parent.name if self.parent else 'unknown',
+            ep.boundary_condition.name,
+            ep.boundary_condition.boundary_condition_object_idf,
+            ep.boundary_condition.sun_exposure_idf,
+            ep.boundary_condition.wind_exposure_idf,
+            ep.boundary_condition.view_factor,
+            len(self.vertices),
+            ',\n\t'.join('%f, %f, %f' % (v[0], v[1], v[2]) for v in self.vertices)
```
chriswmackey: I will add a property to the Face3D class to return vertices that are always lower-left counter-clockwise, which we can use here to make sure that all faces follow global geometry rules.

honeybee_energy/faceutil.py
```diff
@@ -0,0 +1,32 @@
+@property
+def face_to_idf(self):
```
chriswmackey: I am a little bit lost in terms of the relationship of the `Writer` class with the `ladybug_geometry Face3D` class and the overall `honeybee Face` class, which I imagine that the energy properties are extending.

honeybee_energy/properties.py
mostaphaRoudsari: Yes. That is correct.

honeybee_energy/properties.py
mostaphaRoudsari: That's a good suggestion. As discussed this should be possible but we need a strategy for how to implement it with the current structure.

honeybee_energy/faceutil.py
mostaphaRoudsari: Sounds good.

honeybee_energy/faceutil.py
mostaphaRoudsari: `Writer` has not direct relationship with `ladybug_geometry Face3D`.

Writer is used when one calls `Face.to` method to _write_ the face _to_ an another format.

honeybee_energy/properties.py
chriswmackey: Does this mean that you don't plan to address it in this PR and we should just merge this one in?  I could take a stab at trying to implement it if we merge this one in.

honeybee_energy/faceutil.py
chriswmackey: So are the vertices not being stored in a `ladybug_geometry Face3D`?  Are the vertices duplicated on the `Writer` and the `ladybug_geometry Face3D ` class?

mostaphaRoudsari: Vertices are stored in Face itself and are referenced from `Writer`.

honeybee_energy/properties.py
mostaphaRoudsari: It is possible. I have been thinking about the best way to implement it. I'm going to give it a try today and send a PR.

honeybee_energy/properties.py
mostaphaRoudsari: Let me give it a try first today and you can modify it as needed.

mostaphaRoudsari: I changed the boundary condition to return the same instance when possible.
Construction is also set to construction-set structure similar to what you have suggested and documented in UrbanOpt document. All the code is MVP to show the workflow and is missing all the types and boundary conditions.

honeybee_energy/properties.py
mostaphaRoudsari: Now it is initiated without a construction. The construction is set based on a generic construction-set and will be overwritten when faces are added to a zone unless they have been assigned by user. 

PS: I think I need to add zone to show what I mean here.

honeybee_energy/faceutil.py
mostaphaRoudsari: I changed how writer is structured so one need to send the `face` itself to writer and all the properties including vertices are accessed from face itself.

mostaphaRoudsari:
@chriswmackey this is ready for your review.

honeybee_energy/properties.py
chriswmackey: This looks good @mostaphaRoudsari . Thank you.

honeybee_energy/properties.py
chriswmackey: This definitely feels like the right strategy.  I'm not so sure we want to have an `add_face` method on the zone unless we plan to implement the check to ensure that the zone is a closed volume in honeybee core but I agree with this general strategy and we can address this later.

honeybee_energy/faceutil.py
chriswmackey: This seems like the better way to do it even though I understand that it is not as elegant.

chriswmackey:
This looks good @mostaphaRoudsari .  It's a lot better now and I can fill in a lot of the missing info.  I have a few minor comments (for example, I don't think we should support `Zone` boundary conditions now since they are not compatible with future comfort mapping workflows and there are no good ways of auto-solving adjacency with them).  However, I can make these small tweaks once we merge it in and I start filling out the missing properties.

chriswmackey:
Please feel free to do the honors of merging it in.

PR #5, opened 2019-06-03
refactor(boundary condition): move basic boundary conditions to honeybee-core

mostaphaRoudsari:


honeybee_energy/properties.py
```diff
@@ -1,18 +1,20 @@
...
+                 '_construction', '_constructionset')
+
+    def __init__(self, face_type, boundary_condition, construction=None):
+        # private properties to set construction from construction-set.
+        self._face_type = face_type
+        self._boundary_condition = boundary_condition
+        # construction set by user
+        self.construction = construction
+        # Constructionset can only be be set from the parent zone
         self._constructionset = None
```
chriswmackey: I agree that this is the right approach but I just wanted to hear your thoughts about whether this `_constructionset` will be an id or an actual `ConstructionSet` object.  I feel that it could be an object instance and we would still be ok in Grasshopper as long as we add a `__copy__` or `duplicate` method to `EnergyProperties` that doesn't make a copy of the whole `ConstructionSet` instance.  We can have something similar on the zone level and, as long as we make the `ConstructionSet` objects immutable, this should all be good. What do you think?

honeybee_energy/properties.py
```diff
@@ -21,18 +23,16 @@ def construction(self):
...
             # set by user
             return self._construction
         elif self._constructionset:
             # set by parent zone
-            return self._constructionset(self.face_type, self.boundary_condition)
+            return self._constructionset(self._face_type, self._boundary_condition)
         else:
-            # not set yet - use generic construction set
-            return generic_constructionset(self.face_type, self.boundary_condition)
+            return None
```
chriswmackey: Why have we gotten rid of the generic construction set here?

honeybee_energy/boundarycondition.py
```diff
@@ -1,96 +1,10 @@
...
-        _BoundaryCondition.__init__(self)
-        self.boundary_condition_object = other_object
+Note to developers:
+    See _extend_honeybee to see how to add these boundary conditions to
+    honeybee.boundarycondition module.
+"""
+from honeybee.boundarycondition import _BoundaryCondition, Surface
 
 
 class Zone(Surface):
```
chriswmackey: As I said in my earlier comments, I don't think we should support `Zone` boundary conditions now since they are not  compatible with our methods for auto-solving adjacency, calculating closed zone volume, or generating thermal comfort maps.  This is because these `Zone` BCs expect you to specify interior surfaces only once on one of the adjacent zones and then use the `Zone` BC to specify the other adjacent zone.  This means that we will not have have fully closed volume geometries for some zones, creating the issues I note above and generally making it difficult to work with individual zones in isolation of other zones.  We can take it out of this PR now or we could leave it in and I will remove it with my next commit but I just wanted to let you know my thoughts on this.

setup.py
```diff
@@ -16,7 +16,7 @@
     long_description_content_type="text/markdown",
     url="https://github.com/ladybug-tools/honeybee-energy",
     packages=setuptools.find_packages(exclude=["tests"]),
-    install_requires=['git+https://github.com/ladybug-tools/honeybee-core.git'],
+    install_requires=["honeybee-core"],
```
chriswmackey: I didn't realize this could work.  Can I similarly swap out the ladybug core reference to the github URL of ladybug_geometry for `ladybug-geometry`?

honeybee_energy/properties.py
mostaphaRoudsari: I haven't figured that out yet as it depends on how we will handle constructions/ materials. It can be an id if we can retrieve it from a database or what you mentioned above. We will know the best solution once we have the full workflow. I would say for now your solution makes a lot of sense.

honeybee_energy/properties.py
mostaphaRoudsari: Because it was working against us in `to_dict` conversion and will work against us in creating a Model. In both cases passing None is much simpler to evaluate. We know that we must use the Default Construction Set however if we had some default value then we have to do an extra check to see if it is the default construction for this type.

I know it will be confusing in GH interface so we have to find a better solution there to expose default values but as we discussed we will optimize the core libraries for Pollination and find workarounds for plugins.

setup.py
mostaphaRoudsari: Yes and I think that is the right thing to do.

honeybee_energy/boundarycondition.py
mostaphaRoudsari: https://github.com/ladybug-tools/honeybee-energy/pull/5/commits/b2c2b596e57c2bea904434f26f4afeac3a6c7c5f

setup.py
mostaphaRoudsari: I should also add that this will not work currently as deployment of honeybee-core failed but it will work once honeybee-core is available on PyPI.

honeybee_energy/properties.py
chriswmackey: I see. That makes sense and I agree.

PR #8, opened 2019-06-17
feat(material): Adding Initial Energy Materials

chriswmackey:
This pull request contains a complete implementation of all energy constructions and materials presently within the [energy-model-schema](https://github.com/ladybug-tools-in2/energy-model-schema).

In addition to storing material properties and performing checks on inputs, the materials module possess functions for calculating (correct) R-values and U-values.  This is in contrast to honeybee-legacy, which never had correct methods for determining the thermal resistance of window air gaps under different situations. The U-values that we get from window constructions now align with those of LBNL WINDOW to within 0.01 W/m2-K.

The construction objects also possess properties for quickly accessing interior / exterior emissivities and reflectances of constructions, both of which are critical to the implementation of thermal comfort maps.  The quick access to visible reflectance and transmittance will also likely prove useful in workflows of generating Radiance daylight models from energy models.

Currently, this PR lacks checks but I am sending it as a full PR since the general structure is unlikely to change as I add tests and any high-level feedback that can be given earlier is helpful.


honeybee_energy/construction.py
```diff
@@ -29,5 +77,480 @@ def to_idf(self):
...
+        if isinstance(self.materials[0], _EnergyWindowMaterialShadeBase):
+            return 'Exterior'
+        elif isinstance(self.materials[-1], _EnergyWindowMaterialShadeBase):
+            return 'Interior'
+        elif self.has_shade:
+            return 'Between'
+        else:
+            return None
+
+    def temperature_profile(self, outside_temperature=-18, inside_temperature=21):
```
mostaphaRoudsari: This is a very nice method and can be very helpful for visualization too.

honeybee_energy/material/_base.py
```diff
@@ -0,0 +1,91 @@
...
+        self.name = name
+
+    @property
+    def name(self):
+        """Get or set the text string for material name."""
+        return self._name
+
+    @name.setter
+    def name(self, name):
+        self._name = valid_ep_string(name, self._name_descr)
```
mostaphaRoudsari: why not just hard code `'material name'` here? I assume you don't overwrite this method in subclasses so you probably won't reuse `self._name_descr`.

honeybee_energy/material/_base.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class _EnergyMaterialBase(object):
+    """Base energy material.
+
+    Properties:
+        name
+    """
+    _name_descr = 'material name'
+    _is_window_material = None
```
mostaphaRoudsari: should we set this in instance level instead of class level? If one is not careful this can end up being an issue. What is the advantage of having it as a class attribute?

honeybee_energy/material/_base.py
```diff
@@ -0,0 +1,91 @@
...
+        return self.__repr__()
+
+    def __repr__(self):
+        return 'Base Energy Material:\n{}'.format(self.name)
+
+
+class _EnergyMaterialOpaqueBase(_EnergyMaterialBase):
+    """Base energy material for all opaque material types."""
+    _is_window_material = False
+    _rough_types = ('VeryRough', 'Rough', 'MediumRough',
```
mostaphaRoudsari: This looks like a list of constant values to me. You may want to use `ROUGHTYPES` instead.

honeybee_energy/material/_base.py
mostaphaRoudsari: I see similar design pattern in other classes. let's discuss it in one place here.

honeybee_energy/material/glazing.py
```diff
@@ -0,0 +1,439 @@
...
+        u_value
+        r_value
+    """
+    __slots__ = ('_name', '_thickness', '_solar_transmittance', '_solar_reflectance',
+                 '_solar_reflectance_back', '_visible_transmittance',
+                 '_visible_reflectance', '_visible_reflectance_back',
+                 '_infrared_transmittance', '_emissivity', '_emissivity_back',
+                 '_conductivity', '_dirt_correction', '_dirt_correction',
+                 '_solar_diffusing')
+    _thick_descr = 'window material thickness'
```
mostaphaRoudsari: Not a big fan of them being properties. I know it might be more typing but they should be kept inside methods if you ask me. It can also be confusing for other developers on what these attributes are and why are they exposed.

honeybee_energy/properties.py
```diff
@@ -25,19 +25,17 @@ def construction(self):
...
+        elif self._constructionset:  # set by parent zone
             return self._constructionset(self._face_type, self._boundary_condition)
         else:
             return None
 
     @construction.setter
     def construction(self, value):
         if value:
-            assert isinstance(value, Construction), \
+            assert isinstance(value, OpaqueConstruction), \
```
mostaphaRoudsari: I know this is just an update to the existing code but you probably want to do this check against the base construction class and not only OpaqueConstruction.

tests/construction_test.py
```diff
@@ -0,0 +1,75 @@
...
+from honeybee_energy.material.glazing import EnergyWindowMaterialGlazing
+from honeybee_energy.material.gas import EnergyWindowMaterialGas
+
+from honeybee_energy.construction import OpaqueConstruction, WindowConstruction
+
+import unittest
+import pytest
+
+
+class MaterialTestCase(unittest.TestCase):
```
mostaphaRoudsari: Any reason you are going back to using unittest? Not that I have any issues with the approach but is this something that you want to follow for other repos or just a place holder for now.

honeybee_energy/construction.py
```diff
@@ -29,5 +77,480 @@ def to_idf(self):
...
+        """Get a list of temperatures at each material boundary between R-values."""
+        r_factor = sum(r_values)
+        delta_t = inside_temperature - outside_temperature
+        temperatures = [outside_temperature]
+        for i, r_val in enumerate(r_values):
+            temperatures.append(temperatures[i] + delta_t * (r_val / r_factor))
+        return temperatures
+
+    def __copy__(self):
+        return self.__class__(self.name, self.materials)
```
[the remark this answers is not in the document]
chriswmackey: Good point! I will add that now.

honeybee_energy/construction.py
chriswmackey: Glad to hear that you appreciate it!
It was quite a pain in the neck to write correctly for gas layers (which is needed for interior / exterior air films as well as window gaps).  But it has made the u-value calculations of windows much more accurate and I think that you are right that we can produce some good visualizations with it.

honeybee_energy/material/_base.py
chriswmackey: I originally had it that way but I changed it for 2 reasons (one of which is a good one and the other I am not sure about):

1) There are some cases where I reuse the input name since we have some materials with 30 inputs (like blinds) and I don't need to give a different input name for `back_beam_solar_reflectance` and `front_diffuse_solar_reflectcance` (Just `solar_reflectance` is fine). This is the good reason but it doesn't really apply to the `_name_descr` here.

2) The way I put the `_name_descr` variable, I know that it will use the same string for every instance of the class, thereby saving some memory from the case where every class has a new string.  My understanding was that hard-coding a string here will make a new string for each and every instance of the material but I have to admit I was not entirely sure of this. Do you know if my assumption is at all correct?

honeybee_energy/material/_base.py
chriswmackey: It is a hidden attribute that is exposed as a read-only property.  The reason why its not set on the instance is that it needs to be fundamentally tied to the class (it is basically an indicator of what constructions the material class can be a part of).  The main reason why it is there is because I hid the base classes for different types of materials (the base classes do not have enough to be simulate-able and it didn't feel right to expose them). So this property allows users to check the type of material without needing to access the hidden base classes to make `isinstance` calls.

You can see that I don't use this property anywhere in the rest of the API because I know where the hidden classes are and I just use `isinstance`.  However, I feel that it is important to include for the users of the honeybee energy library.

honeybee_energy/material/_base.py
chriswmackey: Ah. Good point. I forgot that convention. Ok I will use that throughout the library.

honeybee_energy/material/glazing.py
chriswmackey: Ok. If the answer to part 2 of my question above is that it does not save memory, I'll get rid of all of them without questions asked.

honeybee_energy/properties.py
chriswmackey: I tried to understand what your plan is in `honeybee-core` for aperture faces but, from what I can tell, they are not there now.  Once you add apertures, I can write this correctly for both opaque and glazing surfaces instead of assuming everything is opaque.
Whatever the case, you should not be able to assign a window construction to an opaque parent surface and vice-versa.  This is something we should catch when it is assigned rather than waiting for OpenStudio / E+ to give us an error.

tests/construction_test.py
chriswmackey: This is what I have done on all of my repos.  If this one is backward, can you point me to an example of "the way forward"?

honeybee_energy/material/glazing.py
chriswmackey: After thinking about this in light of the `__init__` defaults that you mentioned to me the other day, I am nearly positive that the defaults here get stored on the class and not the instance, meaning what I did has no memory savings. Ok, I will keep the string defaults inside the methods.

honeybee_energy/properties.py
mostaphaRoudsari: > Whatever the case, you should not be able to assign a window construction to an opaque parent surface and vice-versa. This is something we should catch when it is assigned rather than waiting for OpenStudio / E+ to give us an error.

This is true and shows a limitation of the current design. Properties class is currently used for all different face types. We might need to subclass Properties class then and assign different ones to different Face types and see how that will work out. I'm going to close this for now.

tests/construction_test.py
mostaphaRoudsari: Just using methods as suggested for pytest? Here is an example:

https://github.com/ladybug-tools/honeybee-radiance/blob/development/tests/plastic_material_test.py

honeybee_energy/material/_base.py
mostaphaRoudsari: For some reason GitHub doesn't show me the full file. I can follow your logic but you can keep it in instance and then expose it as property and it will do the same. As you said in the near future more people should be able to understand the logic and this is one of those cases that being explicit will be probably better.

```python
class _base(object):
    def __init__(self):
        self._is_window_material = False

class WindowMaterial(_base):
    def __init__(self):
         _base.__init__(self)
        # now set to False
        self._is_window_material = False

class OpaqueMaterial(_base):
    def __init__(self):
         _base.__init__(self)
        # It is good as is.
```

chriswmackey: > you can keep it in instance and then expose it as property and it will do the same
... yes except that I will have to add it to `__slots__` and it will take up a bit more memory since every instance will have its own version of this property.  But that's not as big of a reason as the fact that is property is supposed to be a marker of the inheritance tree of objects under the hood. I see that your example is trying to preserve this by always calling `__init__` of the parent class in the child class but it is not clear to developers that, by calling `_base._init()` they need to add the `_is_window_material` property to `__slots__`.
And I still don't understand what the issue is with the code that is there now.  Are you worried about someone overwriting the hidden property that is stored on the parent class?  If so, why is it that you are worried about people overwriting this hidden property in particular and not any of the other hidden properties like `__slots__`?  I think people generally recognize that hidden properties should not be overwritten unless they know exactly what they are doing.

tests/construction_test.py
chriswmackey: I can write the tests that way but I still don't see how they are run. Is there some other file that interacts with this one to call each of the test functions?

chriswmackey:
@mostaphaRoudsari ,
I agree that we need `to_radiance` methods and I would also add that we need `from_ep_string` methods for those people wishing to import current E+ libraries. I think it may also be worth adding `from_standards_gem_dict` to build the material object from the standards gem json dictionaries.  And, lastly, we definitely need `from_dict` class methods, which I still have yet to implement.  I need to put down this PR for the rest of this week to study for my architecture exam but I will add these things first thing next week along with tests for everything.

chriswmackey:
@mostaphaRoudsari ,
Two questions that popped into my mind about how you imagine the `to_radiance` methods working:
1) Should the method simply return a Radiance string or should it return a `honeybee_radiance` material object (in which case , it should be be code-injected from `honeybee_radiance`)?
2) It would be really nice to use `genbsdf` to generate a BSDF material of venetian blinds in the case that a blind material is in the window construction. Do you have any recommendations for how to implement this (ie. should it be written into the xml file and the file path put into the Radiance string)? How are you currently handing external BSDF files in `honeybee_radiance`?

honeybee_energy/material/_base.py
mostaphaRoudsari: >  I see that your example is trying to preserve this by always calling __init__ of the parent class in the child class but it is not clear to developers that, by calling _base._init() they need to add the _is_window_material property to `__slots__`.

They don't have to add it to `__slots__` unless they want to use `__slots__` for the subclass which in that case this is not any different from other attributes. Here is the updated example.

```python
class _base(object):
    __slots__ = ('_is_window_material',)
    def __init__(self):
        self._is_window_material = False

class WindowMaterial(_base):
    __slots__ = ('_is_window_material',)
    def __init__(self):
         _base.__init__(self)
        # now set to False
        self._is_window_material = False

class OpaqueMaterial(_base):
    __slots__ = ()    
    def __init__(self):
         _base.__init__(self)
        # It is good as is.
```

tests/construction_test.py
mostaphaRoudsari: You can run all tests running `python -m pytest .\tests` on Windows. This is how we already run all the tests on travis:

https://github.com/ladybug-tools/honeybee-energy/blob/29688c62d8655ae73bb7a408dfa163ae68c6d0bd/.travis.yml#L11

See pytest test discovery conventions here:

https://docs.pytest.org/en/latest/goodpractices.html#conventions-for-python-test-discovery

mostaphaRoudsari:
Great question! Opened an issue so we can talk about it in-detail: https://github.com/ladybug-tools/honeybee-energy/issues/9

honeybee_energy/material/_base.py
chriswmackey: As discussed, I went with the same setup that you have in `honeybee_radiance`

chriswmackey:
@mostaphaRoudsari ,

All tests are complete and I finished adding in all of the `classmethods` that I mentioned before (`from_dict`, `from_standards_dict` and `from_idf`). Let me know if you want to do a final review before we merge it or we should just merge it now.

chriswmackey:
I am just going to merge it since you already reviewed it before and not that much has changed.

honeybee_energy/construction.py
```diff
@@ -1,33 +1,1111 @@
...
+        materials_dict = cls._idf_materials_dictionary(ep_mat_strings)
+        ep_strs = cls._parse_ep_string(ep_string)
+        materials = [materials_dict[mat] for mat in ep_strs[1:]]
+        return cls(ep_strs[0], materials)
+
+    @classmethod
+    def from_standards_dict(cls, data, data_materials):
+        """Create a OpaqueConstruction from an OpenStudio standards gem dictionary.
+
+        Args:
```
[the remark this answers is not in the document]
chriswmackey: Do you mean that we should make sure that sphinx translates this docstring correctly?

honeybee_energy/construction.py
```diff
@@ -1,33 +1,1111 @@
...
+        for constr in constr_props:
+            try:
+                constr_mats = [materials_dict[mat] for mat in constr[1:]]
+                constructions.append(WindowConstruction(constr[0], constr_mats))
+            except KeyError:
+                pass  # the construction is an opaque construction
+        return constructions, materials
+
+    @staticmethod
+    def _idf_materials_dictionary(ep_mat_strings):
```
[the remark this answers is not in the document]
chriswmackey: Ok. You mean that we should have something like [what you have done here](https://github.com/ladybug-tools/honeybee-radiance/blob/master/honeybee_radiance/parser.py)?  We can eventually do that but `honeybee_energy` is different in that not all objects will be backward translatable from IDF.  All of the complex HVAC stuff will never be translatable from IDF but there is also the fact that, given that we are using schedule ruleset objects for most schedules (which exist in OpenStudio but not E+), writing a parser to generate schedule rulesets from the IDF is going to be tough.  The point that I am making is that I am not sure that we will use our own built-in parser for importing anything other than materials/constructions and geometry from IDF files.  Instead, we will probably rely on OpenStudio to translate from IDF or gbXML and then use the OpenStudio model to build the honeybee model.

honeybee_energy/material/_base.py
```diff
@@ -0,0 +1,121 @@
...
+    @name.setter
+    def name(self, name):
+        self._name = valid_ep_string(name, 'material name')
+
+    def duplicate(self):
+        """Get a copy of this construction."""
+        return self.__copy__()
+
+    @staticmethod
+    def _parse_ep_string(ep_string, test_string=None):
```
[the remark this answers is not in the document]
chriswmackey: That's a much better name.  I made the change now in my local copy and the next PR that I send will include this change.

tests/material_opaque_test.py
```diff
@@ -0,0 +1,235 @@
...
+        concrete.density = -1
+    with pytest.raises(Exception):
+        concrete.specific_heat = -1
+    with pytest.raises(Exception):
+        concrete.roughness = 'Medium'
+    with pytest.raises(Exception):
+        concrete.thermal_absorptance = 2
+    with pytest.raises(Exception):
+        concrete.solar_absorptance = 2
+    with pytest.raises(Exception):
```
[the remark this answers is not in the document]
chriswmackey: I have been using very generic exceptions in my tests because the main thing I care about is that an exception of some form is raised and I sometimes realize later that the type of error that I raise is not the most appropriate.  I am being kind of lazy here, though, and I should be deliberate about the type of error I raise.  Ok, I will include more specific errors in the next PR.

honeybee_energy/construction.py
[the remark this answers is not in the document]
mostaphaRoudsari: Yes. I was talking about having them more accessible as they will be useful for others.

> All of the complex HVAC stuff will never be translatable from IDF but there is also the schedule ruleset objects that we are using for most schedules (which exist in OpenStudio but not E+). This means that writing a parser to generate schedule rulesets from the IDF is going to be tough and the overall point that I am making is that I am not sure that we will use our own built-in IDF parser for importing anything other than materials/constructions and geometry from IDF files. Instead, we will probably rely on OpenStudio to translate from IDF or gbXML and then use the OpenStudio model to build the honeybee model.

Agreed! That is why I wrote "idf parser for **geometries and materials**" in my comment.

Also I'm not suggesting us writing our own parser, that's why I said "in case **one** wants to write an idf parser for geometries and materials at some point it can be accessed directly.".

I just think this is something which will be useful to use to other users and is better to be exposed better through the API.

honeybee_energy/construction.py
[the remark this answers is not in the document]
mostaphaRoudsari: Yes. I just don't know who sphinx deals with a dictionary in docstring.

tests/standards/OpenStudio_Standards_materials.json
```diff
@@ -0,0 +1,15579 @@
...
+    "density": 115.81,
+    "specific_heat": 0.2,
+    "material_standard": "CEC Title24-2013",
+    "material_standard_source": "CEC RJ",
+    "code_category": "Masonry Materials",
+    "code_identifier": "1 Coat Stucco",
+    "notes": "From CEC Title24 2013"
+  },
+  "1/2 in. Gypsum Board": {
+    "name": "1/2 in. Gypsum Board",
```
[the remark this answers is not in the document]
chriswmackey: Yes, the standards gem is full of materials like this one that take full advantage of EnergyPlus's lax rules around acceptable characters for names.  Right now, a lot of these characters are getting stripped out of the names when they get imported as Honeybee materials/constructions but I think that we should support these characters in the case of EnergyPlus materials, constructions, schedules, loads, and all other E+ objects (except geometry).  Stripping out acceptable E+ characters for every single object and storing the original name just does not seem like a very practical or elegant solution (I don't imagine that you store the original name for every Radiance object). 

You can see that this is a major reason why I think it is necessary to have [separate type-checking methods for Radiance and EnergyPlus names](https://github.com/ladybug-tools/honeybee-core/blob/master/honeybee/typing.py#L5-L53) since I don't think we should subject EnergyPlus materials to the same limitations as  Radiance materials (ie. no white spaces) and vice versa (ie. max of 100 characters).  I am planning for the next PR that I send to honeybee-core to have a better version of `valid_ep_string` that accepts all acceptable characters in EnergyPlus names.

For clarification, I think that stripping out all but a few types of characters and storing the original name is a good solution for honeybee-core objects like geometry because these objects get shared between EnergyPlus and Radiance simulations.  But, once we are working with objects that are specific to Radiance or EnergyPlus, the naming conventions of the engine should govern.

honeybee_energy/construction.py
```diff
@@ -1,33 +1,1111 @@
...
+        """Create an OpaqueConstruction from an EnergyPlus IDF text string.
+
+        Args:
+            ep_string: A text string fully describing an EnergyPlus construction.
+            ep_mat_strings: A list of text strings for each of the materials in
+                the construction.
+        """
+        materials_dict = cls._idf_materials_dictionary(ep_mat_strings)
+        ep_strs = cls._parse_ep_string(ep_string)
+        materials = [materials_dict[mat] for mat in ep_strs[1:]]
```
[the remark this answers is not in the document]
chriswmackey: Good catch. I was catching the exception on many of the other similar methods but I had missed this one.

tests/standards/OpenStudio_Standards_materials.json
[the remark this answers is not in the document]
mostaphaRoudsari: Thank you for the detailed explanation. I like the plan. The only place that the name change will happen eventually is when a material / construction is translated to Radiance which in that case it should be fine anyways.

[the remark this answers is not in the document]
chriswmackey: Exactly.  You can see that I realized this as I was writing my response to you and I already pushed a change to my fork to ensure constructions get translated with valid Radiance names:
https://github.com/chriswmackey/honeybee-energy/commit/22f6e69b57933262eecb42a7dced90ca1756caa2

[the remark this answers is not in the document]
mostaphaRoudsari: You actually didn't need to do that. Each modifier takes care of changing the name to a valid radiance name. You can just pass the EP name string as is.

https://github.com/ladybug-tools/honeybee-radiance/blob/093eed49719dbef832b7f58b80c09e09ad4c9c78/honeybee_radiance/primitive/primitivebase.py#L242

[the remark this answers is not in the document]
chriswmackey: Ah, good point. Ok, I will take it out, then. There's no need to run the check twice.

PR #15, opened 2019-07-29
feat(to_dict): Adding to_dict method to the Model object

chriswmackey:
This PR is meant to align with [this change to honeybee-core](https://github.com/ladybug-tools/honeybee-core/pull/13). The tests won't run correctly until that PR has been merged in but I have verified that the tests are running ok locally.

At the time of first posting this PR, I have not yet added from_dict methods to the ModelEnergyProperties and so I am initially submitting it as a draft PR.  Once I have the `from_dict` capability worked out, I will make this a ready-to-review PR.


honeybee_energy/properties/aperture.py
```diff
@@ -69,6 +69,18 @@ def is_construction_set_by_user(self):
...
+    def apply_properties_from_dict(self, abridged_data, constructions):
+        """Apply properties from a ApertureEnergyPropertiesAbridged dictionary.
+
+        Args:
+            abridged_data: A ApertureEnergyPropertiesAbridged dictionary (typically
+                coming from a Model).
+            constructions: A dictionary of constructions with constructions names
+                as keys, which will be used to re-assign constructions.
+        """
+        if 'construction' in abridged_data and abridged_data['construction'] is not None:
```
mostaphaRoudsari: if you want to follow EAFP coding style which might actually fit well in this case you can do:

```python
try:
    const = abridged_data['construction']
except KeyError:
   return
else:
   if const is not None:
        self.construction = const
```

This will also save you from accessing the key 3 times and you will just try one time to get it.

mostaphaRoudsari: Same comment for all the other `apply_properties_from_dict` methods.

honeybee_energy/properties/model.py
```diff
@@ -118,9 +125,9 @@ def check_duplicate_construction_set_names(self, raise_exception=True):
                 duplicate_names.append(con_set.name)
         if duplicate_names != []:
             if raise_exception:
-                raise ValueError('The model has the following duplicated '
-                                 'ConstructionSet names:\n{}'.format(
-                                    '\n'.join(duplicate_names)))
+                raise ValueError(
```
mostaphaRoudsari: I can't comment on the lines before this but as I commented on `honeybee-core` for all the duplicate checks like this you should use `set` for quick access.

chriswmackey: Thanks and good point. I will use `set`

honeybee_energy/properties/aperture.py
chriswmackey: I am not entirely sure that I agree that this is the best approach for the case here.  More often than not, there is not a `'construction'` key in the dictionary since I anticipate it will be much more common for people to leave the Face construction blank and use a `ConstructionSet` instead.  In this sense, the `except KeyError` is what will happen most of the time, which is not really that efficient.

It also just feels non-pythonic to be replacing a single line of code with 6 lines.  I will research EAFP in relation to this case but how certain are you that it applies here?

mostaphaRoudsari: See this:  https://stackoverflow.com/a/11360880/4394669

Also I just realized that there is even a better solution for this case:

```python
const = abridged_data.get('construction')
if const is not None:
    self.construction = const
```

chriswmackey: Yes, that definitely seems a lot more pythonic. I would like to do a little performance testing just to be sure before I implement it but, assuming that it actually helps the performance, I will implement it here and across the from_dict methods of honeybee-core like this one](https://github.com/ladybug-tools/honeybee-core/blob/master/honeybee/model.py#L101)

chriswmackey: So the results of the performance test are in and I am glad I ran it because this does not seem to be a case where the usual EAFP philosophy applies.  Specifically, here is the full test that I ran:

```
import time

face_dict_empty = {'type': 'FaceEnergyPropertiesAbridged'}
face_dict_full = {'type': 'FaceEnergyPropertiesAbridged',
                  'construction': 'Test_Construction'}


def lbyl_test(abridged_data):
    if 'construction' in abridged_data and abridged_data['construction'] is not None:
        return abridged_data['construction']


def eafp_test(abridged_data):
    try:
        const = abridged_data['construction']
    except KeyError:
        pass
    else:
        if const is not None:
            return const


def get_test(abridged_data):
    const = abridged_data.get('construction')
    if const is not None:
        return abridged_data['construction']


st_time = time.perf_counter()
for i in range(100000):
    lbyl_test(face_dict_empty)
print('lbyl time (negative): {}'.format(time.perf_counter() - st_time))

st_time = time.perf_counter()
for i in range(100000):
    lbyl_test(face_dict_full)
print('lbyl time (positive): {}'.format(time.perf_counter() - st_time))

st_time = time.perf_counter()
for i in range(100000):
    eafp_test(face_dict_empty)
print('eafp time (negative): {}'.format(time.perf_counter() - st_time))

st_time = time.perf_counter()
for i in range(100000):
    eafp_test(face_dict_full)
print('eafp time (positive): {}'.format(time.perf_counter() - st_time))

st_time = time.perf_counter()
for i in range(100000):
    get_test(face_dict_empty)
print('get time (negative): {}'.format(time.perf_counter() - st_time))

st_time = time.perf_counter()
for i in range(100000):
    get_test(face_dict_full)
print('get time (positive): {}'.format(time.perf_counter() - st_time))
```
... and here are the results:
```
lbyl time (negative): 0.0244677
lbyl time (positive): 0.0343809
eafp time (negative): 0.0549849
eafp time (positive): 0.0343379
get time (negative): 0.0366309
get time (positive): 0.0393076
```
I am running it with Python 3.6 and so it may also be worth testing it in ironpython.
But, according to these results, what I have currently implemented in this PR is already the most efficient in terms of performance.  It's also the most efficient in terms of number of lines of code.  And, if our previous conversation is any indication, I think what I have now is more readable.  So let me know if you agree that we should leave my code the way that it is for now or if you see any fallacies in the way that I set up my tests above.

Lastly, for the record, I could understand how the EAFP philosophy might be more helpful for a case where you expect a lot of the try/except outcome to be positive or when you have longer dictionaries, which might increase look up times (though I remember a few sources telling me that dictionary lookup times remain fairly constant as you add more keys and it's lists that get longer look ups as the length increases).  In any case, we may not want to be so brazen about applying an EAFP philosophy everywhere and it might be wiser to do performance testing like this.

PR #19, opened 2019-08-14
feat(from_dict): EnergyProperties can be re-serialized for Face, etc.

chriswmackey:
This commit actually includes 3 separate changes but it was difficult to break them up:
1. I have updated the library to work with the new honeybee-core, where aperture type no longer exists and doors have an is_glass property.
2. Now, when someone does face.to_dict() or room.to_dict(), the energy properties can be re-serialized when they take that same dictionary and use it with Face.from_dict() or Room.from_dict().
3. I cleaned up the library of constructions and materials, splitting materials off from constructions and adjusting some default values to that they are coordinated with the default honeybee-radiance materials.


chriswmackey:
@mostaphaRoudsari ,

FYI, this PR also now includes the moving of the generic functions to parse and write IDF strings from the `materials` module to the `writer` module.  You were right that it is very helpful to have these functions all in one place so that a range of objects can make use of them.  Right now, I am using the moved functions in a `schedule` module that I am working on.

honeybee_energy/construction.py
```diff
@@ -10,10 +10,10 @@
     EnergyWindowMaterialGasMixture, EnergyWindowMaterialGasCustom
 from .material.shade import _EnergyWindowMaterialShadeBase, EnergyWindowMaterialShade, \
     EnergyWindowMaterialBlind
-
+from .writer import generate_ep_string, parse_ep_string
```
mostaphaRoudsari: should we use `idf` instead of `ep`. It will be helpful specially when `jdf` will also be available so we know it is idf or jdf even though they are both EP format. > `generate_idf_string`, `parse_idf_string`

honeybee_energy/construction.py
```diff
@@ -1189,3 +1178,201 @@ def _layered_r_value(self, temperatures, r_values_init, emiss,
...
+        """Get a copy of this construction."""
+        return self.__copy__()
+
+    def _to_radiance(self, reflectance):
+        try:
+            from honeybee_radiance.primitive.material.plastic import Plastic
+            from honeybee_radiance.primitive.material.mirror import Mirror
+        except ImportError as e:
+            raise ImportError('honeybee_radiance library must be installed to use '
+                              'to_radiance_solar() method. {}'.format(e))
```
mostaphaRoudsari:
```suggestion
                              'to_radiance_* methods. {}'.format(e))
```

honeybee_energy/lib/_loadconstructions.py
```diff
@@ -0,0 +1,35 @@
...
+
+# empty dictionaries to hold idf-loaded materials and constructions
+_idf_opaque_materials = {}
+_idf_window_materials = {}
+_idf_opaque_constructions = {}
+_idf_window_constructions = {}
+
+
+# load other materials and constructions from user-supplied files
+cur_dir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
```
mostaphaRoudsari: what about using `__file__`?

```suggestion
cur_dir = os.path.dirname(__file__)
```

honeybee_energy/lib/_loadconstructions.py
```diff
@@ -0,0 +1,35 @@
...
+construction_lib = cur_dir + '/idf/constructions/'
```
mostaphaRoudsari: I would use os.path.join in such cases. It should work on Windows like this but os.path.join is cleaner.

```suggestion
construction_lib = os.path.join(cur_dir, 'idf', 'constructions')
```

honeybee_energy/lib/_loadconstructions.py
```diff
@@ -0,0 +1,35 @@
...
+for f in os.listdir(construction_lib):
+    f_path = construction_lib + f
```
mostaphaRoudsari: same here. let's use `os.path.join` to ensure it doen't fail.

honeybee_energy/lib/_loadconstructions.py
```diff
@@ -0,0 +1,35 @@
...
+    if f_path.endswith('.idf') and os.path.isfile(f_path):
```
mostaphaRoudsari: probably you want to check `isfile` first.

honeybee_energy/properties/shade.py
```diff
@@ -51,42 +39,31 @@ def host(self):
...
-            'Shade transmittance cannot be be set while a transmittance ' \
-            'schedule is assigned.'
-        self._transmittance = float_in_range(value, 0, 1, 'shade transmittance')
+    def construction(self):
+        """Get or set a ShadeConstruction for the shade."""
+        if self._construction:  # set by user
+            return self._construction
+        elif not self._host.has_parent:
+            return generic_context
+        elif hasattr(self._host.parent.properties.energy, 'construction_set'):
```
mostaphaRoudsari: The next 3 checks will be probably cleaner if we write them as a single recursive method but this will also work.

honeybee_energy/lib/materials.py
```diff
@@ -0,0 +1,138 @@
...
+from honeybee_energy.material.opaque import EnergyMaterial
+from honeybee_energy.material.glazing import EnergyWindowMaterialGlazing
+from honeybee_energy.material.gas import EnergyWindowMaterialGas
+
+from ._loadconstructions import _idf_opaque_materials, _idf_window_materials
+
+
+# auto-generate materials and constructions if they were not loaded
+# generic opaque materials
+if 'Generic Brick' not in _idf_opaque_materials:
```
mostaphaRoudsari: My suggestion is to put all the default materials that will be used in this method in a single dictionary and then call them from the dictionary in the checks. That will help to being able to see all the default materials in the same place and will editing or reusing them easier in the future.

chriswmackey:
Thanks for reviewing this @mostaphaRoudsari .  I will get to addressing your comments once I finish sending this PR I am working on now for the datetime module of ladybug-core.

honeybee_energy/construction.py
chriswmackey: I agree that calling it `idf` is more appropriate since it makes it clearer these workflows do not engage OpenStudio.  So I will make this change now.  However, I don't know with certainty that they are going to change the name of the IDF file extension when it goes to JSON.  If they do change it, it seems likely that they will call it `epJSON` and not `JDF` as noted by this presentation here: http://web.eecs.utk.edu/~jnew1/presentations/2017_IBPSA_JSON.pdf. All of this is just to say that we may need to figure out what to do with the JSON energyplus format when the time comes.

honeybee_energy/lib/_loadconstructions.py
chriswmackey: Oh, wow, that's so much cleaner.  I see that it also works well for getting the absolute path in IronPython. Ok, I will fix it here and I also have another place to apply this change: https://github.com/ladybug-tools/ladybug-comfort/pull/21

honeybee_energy/lib/_loadconstructions.py
chriswmackey: Makes sense.

honeybee_energy/properties/shade.py
chriswmackey: Excellent suggestion!  Implemented!

honeybee_energy/lib/materials.py
chriswmackey: I don't follow exactly what you mean here since this code is not a method and the default materials are already in a dictionary (the `_idf_opaque_materials`).  However, I have realized that this is likely a place where EAFP will be better performance than LBYL and so I am going to rewrite this code now to reflect that.  Let me know if that is what you meant.

mostaphaRoudsari: That's not what I meant. I mean something like this for all your placeholder materials that you are using in this file.

https://github.com/mostaphaRoudsari/honeybee-radiance/blob/command/honeybee_radiance/modifierset.py#L15

chriswmackey: @mostaphaRoudsari ,
I think that you might not have understood what the `_idf_opaque_materials` dictionary is since it is virtually equivalent to what you point to in that file.  The only difference is that the dictionary is loaded from a .idf file, which users can edit if they ever want to change the honeybee default materials.  The materials.py file where you made this comment is just adding missing keys to this dictionary in the event that someone accidentally deletes them entirely from the .idf file.

chriswmackey:
Thanks for all of the comments @mostaphaRoudsari .  I took all of your suggestions and even the last comment there resulted in an improvement to the code, even if it seems there was a misunderstanding about what the code was doing.

honeybee_energy/lib/materials.py
mostaphaRoudsari: Thanks! I didn't mean it for `_idf_opaque_materials` dictionary but for all the other ones that you generate inside this file using `EnergyMaterial()`. I can send a PR later that shows what I meant. For now it works great and my suggestion was only meant to help the maintenance in the long run. We will see if it will be an issue at all!

PR #20, opened 2019-08-24
feat(schedule): Add several objects for handling schedules

chriswmackey:
@mostaphaRoudsari ,

I'm sending this draft PR so that you are aware this is coming down the pipeline.  Before I have the PR in a merge-able state, there are a few more methods that I need to add to the ScheduleRuleset Class and some more tests.  However, I imagine that nearly all of the 2000 lines of new code is here to stay so there's certainly no harm in getting a head start on the review.


mostaphaRoudsari:
@chriswmackey, Thanks. I think you should break down this PR into 2. The changes in material and construction files should be a separate PR that we can merge in already and then we can have one for the schedules.

PR #21, opened 2019-08-27
feat(schedule): Add objects for schedules and schedule types

chriswmackey:
I have come to realize that the cleanest, most elegant code that is also the easiest to understand involves a clear separation between how Ladybug Tools matches times to values and how EnergyPlus does it. Before, my code was a messy hybrid of E+ and ladybug core conventions but, with this new change the differences between Ladybug Tools and EnergyPlus can be concisely explained with the following phrase:

EnergyPlus uses "time until" and Ladybug Tools uses "time of beginning".

This single phrase summarizes the complexity of everything including E+'s use of 24:00, the fact that the first timestep E+ reports is not 0:00 but ladybug uses 0:00, etc.

Admittedly, this change may initially seem odd to people who are used to only working in E+ but it means that everything is clear once this principle is understood (there's no having to 're-understand` little idiosyncrasies here and there).


chriswmackey:
Thanks for all of the comments @mostaphaRoudsari .  I was able to address many of them directly but there are bunch that required a longer explanation.  Would you mind looking over my responses to the unresolved issues and letting me know if you have any other thoughts before we merge?

mostaphaRoudsari:
@chriswmackey, I added my suggestions. In all cases it is just a suggestion and the current implementation is just fine. Thanks.

mostaphaRoudsari:
@chriswmackey, thank you for addressing all the comments. In case you haven't noticed I have already approved this PR. Feel free to merge it when you get a chance.

chriswmackey:
@mostaphaRoudsari ,
I saw that you has approved it on your first review but your comments were good and I wanted to respond to them.  Also, there are just a few more tests that I want to add and then I'll merge it in.

PR #24, opened 2019-09-06
feat(schedule): Add ScheduleFixedInterval

chriswmackey:
@mostaphaRoudsari ,  I know that this PR includes a few sample CSV schedule files used in testing and so it may seem like there's a lot of new lines but, in reality, the only file that needs review is `honeybee_energy/schedule/fixedinterval.py`


honeybee_energy/schedule/fixedinterval.py
```diff
@@ -0,0 +1,690 @@
...
+        """Get the integer for the schedule's number of steps per hour."""
+        return self._timestep
+
+    @property
+    def start_date(self):
+        """Get the ladybug Date object noting when the the input values take effect."""
+        return self._start_date
+
+    @property
+    def end_date_time(self):
```
mostaphaRoudsari: I'm sure you have already thought this through but to make sure I understand why this is set-up the way that it is. Does this mean that we can support cases from 21 December to 21 March next year? I _feel_ the end date should be an input by user and we validate start, end and values. This gives a better flexibility but it can also open up cases the the user inputs the wrong values and gets an end date which is not expected without noticing.

honeybee_energy/schedule/fixedinterval.py
```diff
@@ -0,0 +1,690 @@
...
+        """Get a boolean noting whether the schedule is over a leap year.
+
+        Note that this property originates from the leap_year property on the
+        input start_date.
+        """
+        return self._start_date.leap_year
+
+    @property
+    def data_collection(self):
+        """A DataCollection of schedule values at the native start_date and timestep."""
```
mostaphaRoudsari: what is a _native_ start date?

honeybee_energy/schedule/fixedinterval.py
```diff
@@ -0,0 +1,690 @@
...
+                                  self.is_leap_year)
+        data_type, unit = self._get_lb_data_type_and_unit()
+        header = Header(data_type, unit, a_period, metadata={'schedule': self.name})
+        return HourlyContinuousCollection(header, self._values)
+
+    def values_at_timestep(
+            self, timestep=1, start_date=None, end_date=None):
+        """Get a list of sequential schedule values over the year at a given timestep.
+
+        Note that there are two possible ways that these values can be mapped to
```
mostaphaRoudsari: I would at least name the two here. And if possible mention how does it make a difference for the developer. Is it an input to switch between the two? Is there one that is already in use? The description right now just makes me less confident to use this method.

honeybee_energy/schedule/fixedinterval.py
```diff
@@ -0,0 +1,690 @@
...
+            {
+            "type": 'ScheduleFixedInterval',
+            "name": 'Automated Awning Transmittance',
+            "values": [], // list of numbers for the values of the schedule
+            "schedule_type_limit": {}, // ScheduleTypeLimit dictionary representation
+            "timestep": 1, // Integer for the timestep of the schedule
+            "start_date": {}, // Date dictionary representation
+            "out_of_range_value": 0, // Number for the values out of range
+            "interpolate": False // Boolean noting whether to interpolate between values
+            }
```
mostaphaRoudsari: we will see once this is merged but I think you need an extra empty line after the end of the code-block.

honeybee_energy/schedule/fixedinterval.py
```diff
@@ -0,0 +1,690 @@
...
+    def from_dict(cls, data):
+        """Create a ScheduleFixedInterval from a dictionary.
+
+        Note that the dictionary must be a non-abridged version for this
+        classmethod to work.
+
+        Args:
+            data: ScheduleFixedInterval dictionary following the format below.
+
+        .. code-block:: shell
```
mostaphaRoudsari: I would use javascript in this case with a JSON object with // comments.

mostaphaRoudsari:
Also the tests are failing in Python 2 in case you haven't seen that. Let's make sure to fix those before merging this in.

honeybee_energy/schedule/fixedinterval.py
```diff
@@ -0,0 +1,690 @@
...
+                values correspond to.  For example, if each value represents 30
+                minutes, the timestep is 2. For 15 minutes, it is 4. Default is 1,
+                meaning each value represents a single hour. Must be one of the
+                following: (1, 2, 3, 4, 5, 6, 10, 12, 15, 20, 30, 60).
+            start_date: A ladybug Date object to note when the the input values begin
+                to take effect. Default is 1 Jan for a non-leap year. Note that this
+                default usually should not be changed unless you plan to run a
+                simulation that is much shorter than a year and/or you plan to run
+                the simulation for a leap year.
+            out_of_range_value: A value that will be used for all times not covered
```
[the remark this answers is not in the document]
chriswmackey: I was matching the naming convention of OpenStudio for this property, which I think is actually pretty clear that it's the value used for days out of the date range (`default_value` definitely feel more abstract and vague). place_holder_value is a little clearer. Let me think.

honeybee_energy/schedule/fixedinterval.py
```diff
@@ -0,0 +1,690 @@
...
+                by the input values. Typically, your simulation should not need to
+                use this value if the input values completely cover the simulation
+                period. However, a default value may still be necessary for EnergyPlus
+                to run. Default: 0.
+            interpolate: Boolean to note whether values in between intervals should be
+                linearly interpolated or whether successive values should take effect
+                immediately upon the beginning time corrsponding to them. Default: False
```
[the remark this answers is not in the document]
chriswmackey: This is something we should definitely add to the [list of things we want to clean up in the documentation](https://github.com/ladybug-tools/ladybug/issues/158). I don't have a preference for either standard as long as there is a standard. Is there a benefit for going with the Click convention?

honeybee_energy/schedule/fixedinterval.py
chriswmackey: Yes, I have made sure that everything works right now for schedules that used "reversed analysis_periods" as we call them in ladybug. Adding an `end_date_time` input seems unnecessarily annoying, especially when someone just wants to create a schedule for a single-day simulation and the default value is likely to be wrong. If you want, I could add a method that validates that the `end_date_time` is what someone expects but it will literally be `return self.end_date_time == test_end_date_time` so I'm not sure it's really necessary. Maybe I don't understand what you mean by it adding more "flexibility" since it only seems like it will be more limiting and create more work for the user as I see it now.

honeybee_energy/schedule/fixedinterval.py
chriswmackey: Its the `start_date` assigned to the schedule (as opposed to a user-input `start_date`. I can make this clearer in the docstring.

honeybee_energy/schedule/fixedinterval.py
chriswmackey: Ok. I added a more detailed description according to your suggestion.

honeybee_energy/schedule/fixedinterval.py
chriswmackey: Ah, Ok. I added the extra line break.

chriswmackey: Actaully, I just checked my old code where I had done this and the extra line break is unnecessary:
https://www.ladybug.tools/honeybee-energy/docs/honeybee_energy.schedule.day.html

honeybee_energy/schedule/fixedinterval.py
chriswmackey: I'm afraid that I don't follow your comment here.  I use // comments in the dictionary. Where did you see me not use them?

chriswmackey:
Thanks for the comments and the approval, @mostaphaRoudsari .  I fixed the Python 2 failure and I was able to address most of your comments but I feel that I didn't understand some of them and it would be good if you could elaborate a bit before I merge this in.

honeybee_energy/schedule/fixedinterval.py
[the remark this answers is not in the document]
chriswmackey: Thinking about it more, I think you're right that `placeholder_value` is clearer.  I will change it to that.

AntoineDao:
If the question you're asking is whether we should write example/demo scripts using jupyter notebooks I think it's a great idea. Notebooks allow you to add rich text commentary and display plots + graphs as a result of your code execution. We can also use [Binder](https://mybinder.org/) to live host the notebooks for non savvy python users to test code. This feels like a conversation for somewhere else though. I'll ping it to Slack and we can keep it going there.

AntoineDao:
Also, nice PR @chriswmackey 😄

chriswmackey:
To respond about the samples, I fully support going in that direction but I probably won't have time to put together some good examples myself until after Nov 8th. I'll read the article you sent and, if you want to take a stab at developing a standard for us to use for this, I can give feedback on it and follow it.

honeybee_energy/schedule/fixedinterval.py
mostaphaRoudsari: By "This gives a better flexibility" I meant your current solution gives more flexibility but yes it can stay as is.

honeybee_energy/schedule/fixedinterval.py
mostaphaRoudsari: I meant a `.. code-block:: json` will probably give you a better highlight than `.. code-block:: shell`.

I usually use `.. code-block:: python` and then use # for comments as they are python dictionaries but JSON will also work.

chriswmackey: Ah, that makes sense. Ok. Will do.

chriswmackey: Done!

honeybee_energy/schedule/fixedinterval.py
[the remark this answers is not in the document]
mostaphaRoudsari: Cool! We will get to it when we address that issue.

PR #27, opened 2019-09-16
feat(programtype): Add complete object for ProgramType

chriswmackey:
This commit adds the rest of objects that are needed to create a fully-simulate-able energy model. This includes:
* An object to store the `Setpoints `of Rooms and `ProgramTypes`
* All of the properties to the `ProgramType `needed to fully-describe a space type as it exists in the OpenStudio standards gem (the only exception to this is the hot water system)
* An `IdealAirSystem `object to store properties of simple HVAC systems
* A number of properties to both `Rooms `and `Models `to store program-related information

The dictionary output from the `Model.to_dict()` should now describe a simulate-able model. The only things that are needed to enable simulate-able `Model.to.idf()` workflows is to:
* add the methods to `writer.py` that call the `to_idf `methods that currently exist on nearly all objects in this library.
* add a module to store simulation parameters like the run period, north, etc.

@mostaphaRoudsari ,
I apologize that this PR is so long but, when I was this close to putting the keystone in the arch, it seemed to make the most sense to finish everything and write the system-wide tests in addition to just unit tests. At this point, I don't plan on adding any new "real" code to this PR and so it is ready for review.  I still have to write some more tests, though, and I think I will add a generic office `ProgramType `to the `lib`, which will be useful for cases where people want to run tests with a generic type of program.


honeybee_energy/load/lighting.py
```diff
@@ -102,9 +105,12 @@ def radiant_fraction(self):
...
     def radiant_fraction(self, value):
-        self._radiant_fraction = float_in_range(
-            value, 0.0, 1.0, 'lighting radiant fraction')
-        self._check_fractions()
+        if value is not None:
+            self._radiant_fraction = float_in_range(
+                value, 0.0, 1.0, 'lighting radiant fraction')
+            self._check_fractions()
+        else:
+            self._radiant_fraction = 0.32
```
mostaphaRoudsari: This can end up creating strange behavior since you are not checking the fraction afterwards. Is there a reason that we hard-core the value here and also a reason for not calling `check_fractions`?

The same question about other fields with none 0 values.

honeybee_energy/programtype.py
```diff
@@ -63,5 +82,567 @@ def name(self):
...
+            sched.append(self.setpoint.cooling_schedule)
+            if self.setpoint.humidifying_schedule is not None:
+                sched.append(self.setpoint.humidifying_schedule)
+                sched.append(self.setpoint.dehumidifying_schedule)
+        return sched
+
+    @property
+    def unique_schedules(self):
+        """List of all unique schedules contained within the ProgramType."""
+        return list(set(self.schedules))
```
mostaphaRoudsari: I would say the `schedules` property should return the unique list of schedules. Do you see a use case to get the list with duplicate schedules?

honeybee_energy/programtype.py
```diff
@@ -63,5 +82,567 @@ def name(self):
...
+        infiltration = None
+        ventilation = None
+        setpoint = None
+
+        if data['occupancy_schedule'] is not None:
+            occ_sched = ScheduleRuleset.from_standards_dict(
+                data_schedules[data['occupancy_schedule']])
+            act_sched = ScheduleRuleset.from_standards_dict(
+                data_schedules[data['occupancy_activity_schedule']])
+            occ_density = data['occupancy_per_area'] / 92.903
```
mostaphaRoudsari: Is there a plan for OpenStudio standards to have a SI version? I'm asking since it might make sense to add an input for units to be set to SI and / or IP or we can probably do that when the SI version will be available if ever!

honeybee_energy/programtype.py
```diff
@@ -63,5 +82,567 @@ def name(self):
...
+        if self.ventilation is not None:
+            self.ventilation._locked = True
+        if self.setpoint is not None:
+            self.setpoint._locked = True
+
+    def unlock(self):
+        """The unlock() method will also unlock the loads."""
+        self._locked = False
+        if self.people is not None:
+            self.people._locked = False
```
mostaphaRoudsari: calling the private methods outside the class itself is not good practice. I have seen better links but this can do (https://www.oreilly.com/ideas/how-to-make-mistakes-in-python). Search for `private`. This can result in bugs if at some point one changes `unlock` method and does more than just changing `_locked` to False.

honeybee_energy/properties/model.py
```diff
@@ -27,15 +28,17 @@ class ModelEnergyProperties(object):
...
-        unique_materials
-        unique_constructions
-        unique_face_constructions
-        unique_shade_constructions
-        unique_construction_sets
-        global_construction_set
-        unique_schedule_type_limit
-        unique_schedules
-        unique_shade_schedules
+        * unique_materials
```
mostaphaRoudsari: I know this has been there from the start and I should have commented about this probably sooner but it feels that unique is redundant here as these output must be unique anyways. Also in case we want to keep unique in the name then having it at the end of the name is the recommended approach (e.g. materials_unique). I can't find the link in the documentation right now but I remember part of the logic was better readability in the API. Now you will get `unique_` before what really matters (e.g. materials, construction, etc).

honeybee_energy/load/equipment.py
```diff
@@ -318,6 +338,32 @@ def from_dict(cls, data):
         sched, rad_f, lat_f, lost_f = cls._extract_dict_props(data, 'ElectricEquipment')
         return cls(data['name'], data['watts_per_area'], sched, rad_f, lat_f, lost_f)
 
+    @classmethod
+    def from_dict_abridged(cls, data, schedule_dict):
```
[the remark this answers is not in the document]
chriswmackey: I had thought about that but the dictionary of schedules is required when someone passes an `EquipmentAbridged` dictionary but it should definitely not be included when passing an `Equipment` dictionary.  So it becomes vague what inputs are needed when they are squashed into one method.  Having separate methods makes it clear what is required under each situation. Given the way that I have written it, it's also the exact same amount of code either way so I don't see what we gain by having them as one method. I only really see the loss of clarity right now.

honeybee_energy/load/lighting.py
chriswmackey: Very good catch.  I forgot that this is only appropriate when the default is 0.

honeybee_energy/programtype.py
chriswmackey: There were two reasons why I have separate properties for both schedules and unique schedule:
1. It parallels the `ConstructionSet`, which has properties for both `constructions`, which returns a list of all constructions where you know what type of construction it is by its index, and `unique_constructions`, which only gives you the unique ones and where the index doesn't tell you anything (but this is needed for the Model).
2, You are right that it's rare to get the same types of schedules within a single ProgramType (it's far more common to have shared schedules between program types). For this reason, I don't usually want to go through the whole process of hashing the schedule objects to create a python set if I don't need to (this hashing can definitely add up over large models).  However, there are definitely cases where the same schedule is used for two fields (like occupancy and ventilation) and so I need a `unique_schedules` object to make sure that I only pull the unique ones for these cases.

honeybee_energy/programtype.py
chriswmackey: That would be nice. The sense I had gotten from Dan a while back is that it's not really in the plans right now (or it's not in the plans for as long as the spreadsheet lives since it seems the people adding/maintaining the US standards are IP people).  Matt and Marely will know better and I will ask when I get the chance to see them.  I'll add it if they think it's in the works.

BTW, I am thinking that all of these `from_standards_dict` methods will get pulled out of the `honeybee_energy` library and put into the `honeybee_standards` library once we set it up. These methods aren't really practical unless you have the standards gem data present so I feel like it makes sense for `honeybee_standards` to extend the `honeybee_energy` objects with these methods.

honeybee_energy/programtype.py
chriswmackey: Fair point.  I'll go through and fix all of the places where I have custom `lock` and `unlock` methods that are set up like this and have them use the public method instead. For clarification, though, I would draw a distinction between "calling a private method" and "accessing a private attribute" since `_locked` is not a method.   I would say that you're right that you should never set a private variable outside of the module where it is introduced (so what I'm doing here is wrong). However, checking what a private attribute is (in some cases) very helpful and is often necessary when the hidden attribute can differ from the public one.

honeybee_energy/properties/model.py
chriswmackey: Fair point.  Since all of these properties on the model are unique, I think you're right that we can just shorten all of these names.  However, both the `ConstructionSet` and `ProgramType` have separate properties for unique vs. non-unique.  So would you recommend that I change those two cases to follow what you recommend here (that they all have `unique` at the end)?

honeybee_energy/load/equipment.py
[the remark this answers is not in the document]
mostaphaRoudsari: I didn't notice the difference in input which makes sense.

honeybee_energy/programtype.py
mostaphaRoudsari: > However, checking what a private attribute is (in some cases) very helpful and is often necessary when the hidden attribute can differ from the public one.

This is the original link that I was looking for. See their suggestion for resolving such cases: https://docs.quantifiedcode.com/python-anti-patterns/correctness/accessing_a_protected_member_from_outside_the_class.html

I agree that there might be very specific cases but generally speaking I would say access outside the _class_ is not a good practice.

honeybee_energy/properties/model.py
chriswmackey: I removed `unique` from all of the model properties.  Thinking it over, you are right that `_unique` makes more sense at the end of the property.  I'll do this for `ProgramType` and `ConstructionSet`.

mostaphaRoudsari: I tend to say yes but it depends on how much work it will be. It looks like a global search and replace problem to me.

honeybee_energy/programtype.py
mostaphaRoudsari: I can't really follow your logic here but I trust your judgment on the topic. Let's keep both.

honeybee_energy/properties/model.py
chriswmackey: I just implemented it now and it was not much time.

chriswmackey:
Thanks for the quick review @mostaphaRoudsari .  I see that you didn't approve it yet so let me know if you are still working your way through it.  In the mean time, I will finish writing these tests.

PR #30, opened 2019-09-26
feat(construction): Split up construction module into several

chriswmackey:
I have come to realize that the existing construction module was already 1400 lines long and we are far from finished adding to it.  I'll probably implement window frame objects as part of the WindowConstruction and I think DynamicWindowConstruction is going to be it's own object, which holds ShadeControl objects. Given all of this, it will be a lot easier to maintain if we split up the construction module now before we build more on top of it.

@mostaphaRoudsari .  There's no real code change with this PR and I just requested your review so that you are aware of it and to check whether you agree the module should be split up


chriswmackey:
Thanks @mostaphaRoudsari . I was going back and forth about what tag to use and I guess that I was just wary of what the `refactor` tag does to the semantic release counter. I will re-label it upon merging in.

PR #32, opened 2019-10-06
Improve deploy.sh to latest standard

Yash-Garg:


chriswmackey:
I appreciate the effort here, @Yash-Garg but what you have done is not at the standard that we currently use across the new repos. See the [ladybug-tools-template repo](https://github.com/ladybug-tools/ladybug-tools-template/tree/master/%7B%7Bcookiecutter.project_slug%7D%7D) for what we are doing now.  While you would be right to point out that this repo is not up to the standard, this upgrade is already being addressed as a part of [this PR](https://github.com/ladybug-tools/honeybee-energy/pull/31).  I would encourage you to look at our [hacktoberfest board](https://github.com/orgs/ladybug-tools/projects/2) to see the things that we definitely need help on.

PR #34, opened 2019-10-15
feat(output): Add module to handle simulation outputs

chriswmackey:
This commit adds a module to facilitate the requesting of outputs from the simulation.

I still have to add a few more tests but the code is all ready for review, @mostaphaRoudsari .

### DIFF
--- a/honeybee_energy/simulation/output.py
+++ b/honeybee_energy/simulation/output.py
@@ -1,0 +1,533 @@
+# coding=utf-8
+"""Object to hold EnergyPlus simulation outputs."""
+from __future__ import division
+
+from ..reader import parse_idf_string
+from ..writer import generate_idf_string
+
+
+class SimulationOutput(object):
+    """Object to hold EnergyPlus simulation outputs.
+
+    Properties:
+        * outputs
+        * reporting_frequency
+        * include_sqlite
+        * include_html
+        * summary_reports
+    """
+    __slots__ = ('_outputs', '_reporting_frequency', '_include_sqlite',
+                 '_include_html', '_summary_reports')
+    REPORTING_FREQUENCIES = ('Annual', 'Monthly', 'Daily', 'Hourly', 'Timestep')
+
+    def __init__(self, outputs=None, reporting_frequency='Hourly', include_sqlite=True,
+                 include_html=False, summary_reports=['AllSummary']):
+        """Initialize SimulationOutput.
+
+        Args:
+            outputs: A list of EnergyPlus output names as strings, which are requested
+                from the simulation. If None, no outputs will be requested.
+                Note that this object does not check whether the outputs exist
+                within the EnergyPlus IDD or are request-able from a given Model.
+                (eg. ['Zone Ideal Loads Supply Air Total Cooling Energy']).
+                Default: None.
+            reporting_frequency: Text for the frequency at which the outputs
+                are reported. Default: 'Hourly'.
+                Choose from the following:
+                    * Annual
+                    * Monthly
+                    * Daily
+                    * Hourly
+                    * Timestep
+            include_sqlite: Boolean to note whether a SQLite report should be
+                generated from the simulation, which contains all of the outputs and
+                summary_reports. Default: True.
+            include_html: Boolean to note whether an HTML report should be generated
+                from the simulation, which contains all of the summary_reports.
+                Default: False.
+            summary_reports: A list of EnergyPlus summary report names as strings.
+                An empty list or None will result in no summary reports.
+                Default: ['AllSummary']. See the Input Output Reference SummaryReports
+                section for a full list of all reports that can be requested.
+                (https://bigladdersoftware.com/epx/docs/9-1/input-output-reference/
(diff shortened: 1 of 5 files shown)


honeybee_energy/simulation/output.py
```diff
@@ -0,0 +1,518 @@
...
+        if output_variables is not None:
+            outputs = []
+            for out_str in output_variables:
+                ep_out_str = parse_idf_string(out_str, 'Output:Variable,')
+                outputs.append(ep_out_str[1])
+                try:
+                    frequency = ep_out_str[2] if ep_out_str[2] != '' else 'Hourly'
+                except IndexError:
+                    pass  # shorter output variable with default hourly frequencey
+        print(frequency)
```
mostaphaRoudsari: Is this intentional?

honeybee_energy/simulation/output.py
```diff
@@ -0,0 +1,518 @@
...
+            sqlite: An IDF Output:SQLite string to request the SQLite file from
+                the simulation. Will be None if include_sqlite is False.
+            surfaces_list: An IDF Output:Surfaces:List string to ensure surface
+                information is written into the ultimate .eio file.
+        """
+        style = 'CommaAndHTML' if self.include_html else 'Comma'
+        table_style = generate_idf_string(
+            'OutputControl:Table:Style',
+            (style, 'JtoKWH'), ('column separator', 'unit conversion'))
+        output_variables = [self._output_to_idf(out_p) for out_p in self._outputs] if \
```
mostaphaRoudsari: I would use `set(self._outputs)` instead to remove possible duplicate outputs.

```suggestion
        output_variables = [self._output_to_idf(out_p) for out_p in set(self._outputs)] if \
```

honeybee_energy/simulation/output.py
```diff
@@ -0,0 +1,518 @@
...
+
+        These reports will be requested from the simulation and, if None,
+        no summary reports will be written.
+        """
+        return tuple(self._summary_reports)
+
+    @summary_reports.setter
+    def summary_reports(self, value):
+        if value is not None:
+            if not isinstance(value, list):
```
mostaphaRoudsari: This will fail for tuple or iterator inputs. A better test is probably checking against being string or bytes.

honeybee_energy/simulation/output.py
```diff
@@ -0,0 +1,518 @@
...
+    REPORTING_FREQUENCIES = ('Annual', 'Monthly', 'Daily', 'Hourly', 'Timestep')
+
+    def __init__(self, outputs=None, reporting_frequency='Hourly', include_sqlite=True,
+                 include_html=False, summary_reports=['AllSummary']):
+        """Initialize SimulationOutput.
+
+        Args:
+            outputs: A list of EnergyPlus output names as strings, which are requested
+                from the simulation. If None, no outputs will be requested.
+                Note that this object does not check whether the outputs exist
```
mostaphaRoudsari: I agree that we don't need this check for now but this might worth opening an issue for later development to check the outputs against the EnergyPlus IDD.

chriswmackey:
Thanks for the lightning-fast review.  I'll see if I can address them quickly now.

honeybee_energy/simulation/output.py
chriswmackey: Nope! Good catch!

honeybee_energy/simulation/output.py
chriswmackey: Let me think about this and possibly implement something tomorrow. If we are going to use sets, I'm not sure this is really the right place to do it.

honeybee_energy/simulation/output.py
chriswmackey: I don't know what you mean and you might be misunderstanding what this code is doing. If someone inputs a tuple or an iterator, I want it to be cast to a list. It's not going to fail in this case.

honeybee_energy/simulation/output.py
chriswmackey: Ok, I added the issue: https://github.com/ladybug-tools/honeybee-energy/issues/35

honeybee_energy/simulation/output.py
mostaphaRoudsari: Sorry. You are right. I misread the code. But now this code fails if one inputs a string. This check will create a list of characters.

chriswmackey: Ah, that's a good point. I always forget strings are also iterables and, particularly in this case, it will result in a very common error where someone passes a string of a single E+ output. Ok, I'll add an extra check in here.

honeybee_energy/simulation/output.py
chriswmackey: After thinking about it a bit, I decided it was best to use a `set` as the hidden attribute for `_outputs` instead of a `list`. This way, we avoid duplicate output names throughout the process of using the object rather than just waiting till it's written to an IDF or a dictionary to get rid of possible duplicates. The public attribute `outputs` is still a tuple so nothing has really changed on the interface end.

PR #36, opened 2019-10-18
feat(writer): Add writer objects to go to.idf

chriswmackey:
Working with the new structure of honeybee-core writers. This commit officially brings everything together to make a simulate-able IDF!

### DIFF
--- a/honeybee_energy/properties/model.py
+++ b/honeybee_energy/properties/model.py
@@ -18,6 +18,7 @@ from ..programtype import ProgramType
 from ..schedule.typelimit import ScheduleTypeLimit
 from ..schedule.ruleset import ScheduleRuleset
 from ..schedule.fixedinterval import ScheduleFixedInterval
+from ..writer import generate_idf_string
 
 try:
     from itertools import izip as zip  # python 2
@@ -260,6 +261,33 @@ class ModelEnergyProperties(object):
                                                program_types):
                     program_types.append(room.properties.energy._program_type)
         return list(set(program_types))  # catch equivalent program types
+
+    def building_idf(self, solar_distribution='FullInteriorAndExteriorWithReflections'):
+        """Get an IDF string for Building that this model represents.
+
+        Args:
+            solar_distribution: Text desribing how EnergyPlus should treat beam solar
+                radiation reflected from surfaces. Default:
+                FullInteriorAndExteriorWithReflections. Choose from the following:
+                    * MinimalShadowing
+                    * FullExterior
+                    * FullInteriorAndExterior
+                    * FullExteriorWithReflections
+                    * FullInteriorAndExteriorWithReflections
+        """
+        values = (self.host.name,
+                  self.host.north_angle,
+                  self.terrain_type,
+                  '',
+                  '',
+                  solar_distribution)
+        comments = ('name',
+                    'north axis',
+                    'terrain',
+                    'loads convergence tolerance',
+                    'temperature convergence tolerance',
+                    'solar distribution')
+        return generate_idf_string('Building', values, comments)
 
     def check_duplicate_material_names(self, raise_exception=True):
         """Check that there are no duplicate Material names in the model."""
--- a/honeybee_energy/schedule/day.py
+++ b/honeybee_energy/schedule/day.py
@@ -425,9 +425,16 @@ class ScheduleDay(object):
 
         return cls(data['name'], data['values'], times, interpolate)
 
-    def to_idf(self):
-        """IDF string representation of ScheduleDay object."""
-        fields = [self.name, '']
+    def to_idf(self, schedule_type_limit=None):
+        """IDF string representation of ScheduleDay object.
+
+        Args:
+            schedule_type_limits: Optional ScheduleTypeLimit object, which will
+                be written into the IDFstring in order to validate the values
+                within the schedule during the EnergyPlus run.
+        """
+        fields = [self.name, ''] if schedule_type_limit is None else \
+            [self.name, schedule_type_limit.name]
         fields.append('No' if not self.interpolate else 'Linear')
         comments = ['schedule name', 'schedule type limits', 'interpolate to timestep']
         for i in range(len(self._values)):
--- a/honeybee_energy/simulationparameter.py
+++ b/honeybee_energy/simulationparameter.py
@@ -25,6 +25,7 @@ class SimulationParameter(object):
         * simulation_control
         * shadow_calculation
         * sizing_parameter
+        * global_geometry_rules
     """
     __slots__ = ('_output', '_run_period', '_timestep', '_simulation_control',
                  '_shadow_calculation', '_sizing_parameter')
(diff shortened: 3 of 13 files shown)


chriswmackey:
Sorry for pushing changes so many times.  This is officially ready to be reviewed/merged and I can say (AT LAST!) that IDFs produced by this code have been run through EnergyPlus!

honeybee_energy/simulationparameter.py
```diff
@@ -286,22 +298,42 @@ def to_idf(self):
...
-        shadow_calc_str = self.shadow_calculation.to_idf()
-        sizing_par_str = self.sizing_parameter.to_idf()
-
-        return '/n/n'.join([table_style, output_vars_str, reports, sqlite, surfaces,
-                            header_str, sim_control_str, shadow_calc_str, timestep_str,
-                            run_period_str, holiday_str, daylight_saving_time_str,
-                            sizing_par_str])
+        sim_param_str.append(run_period_str)
+        if holidays is not None:
+            sim_param_str.append('\n\n'.join(holidays))
```
mostaphaRoudsari: Wouldn't '\n' just do it in this case?

honeybee_energy/writer.py
```diff
@@ -22,44 +28,368 @@ def generate_idf_string(object_type, values, comments=None):
         ep_str = object_type + ',\n ' + '\n '.join(
             '{},{}!- {}'.format(val, space, com) for val, space, com in
             zip(values[:-1], spaces[:-1], comments[:-1]))
-        ep_str = ep_str + '\n {};{}!- {}'.format(
-            values[-1], spaces[-1], comments[-1])
+        if len(values) == 1:
+            ep_str = ep_str + '{};{}!- {}'.format(values[-1], spaces[-1], comments[-1])
```
mostaphaRoudsari: This is not that much of text concatenate but if you look online people have strong opinion about concatenating strings in Python.

honeybee_energy/writer.py
```diff
@@ -22,44 +28,368 @@ def generate_idf_string(object_type, values, comments=None):
...
+    values = (door.name,
+              'Door' if not door.is_glass else 'GlassDoor',
+              door.properties.energy.construction.name,
+              door.parent.name if door.has_parent else 'unknown',
+              door_bc_obj,
+              door.boundary_condition.view_factor,
+              '',  # TODO: Implement Frame and Divider objects on WindowConstructions
+              '1',
+              len(door.vertices),
+              ',\n '.join('%.3f, %.3f, %.3f' % (v[0], v[1], v[2])
```
mostaphaRoudsari: What about unpacking v?

```suggestion
              ',\n '.join('%.3f, %.3f, %.3f' % (*v)
```

mostaphaRoudsari: I see this in a couple of places but it is not really critical.

honeybee_energy/writer.py
```diff
@@ -22,44 +28,368 @@ def generate_idf_string(object_type, values, comments=None):
...
+    """
+    # create the Shading:Detailed IDF string
+    trans_sched = shade.properties.energy.transmittance_schedule.name if \
+        shade.properties.energy.transmittance_schedule is not None else ''
+    if shade.has_parent and not isinstance(shade.parent, Room):
+        if isinstance(shade.parent, Face):
+            base_srf = shade.parent.name
+        else:  # Aperture for parent
+            try:
+                base_srf = shade.parent.parent.name
```
mostaphaRoudsari: In what case this would / should work?

honeybee_energy/writer.py
```diff
@@ -22,44 +28,368 @@ def generate_idf_string(object_type, values, comments=None):
...
+
+def _instance_in_array(object_instance, object_array):
+    """Check if a specific object instance is already in an array.
+
+    This can be much faster than  `if object_instance in object_arrary`
+    when you expect to be testing a lot of the same instance of an object for
+    inclusion in an array since the builtin method uses an == operator to
+    test inclusion.
+    """
+    for val in object_array:
```
mostaphaRoudsari: If you are using it too many times than it might sense to create a `set` from `object_instance` and do all your inclusion checks.

honeybee_energy/properties/model.py
```diff
@@ -261,6 +262,33 @@ def program_types(self):
                     program_types.append(room.properties.energy._program_type)
         return list(set(program_types))  # catch equivalent program types
 
+    def building_idf(self, solar_distribution='FullInteriorAndExteriorWithReflections'):
```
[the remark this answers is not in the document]
chriswmackey: What do you mean? Do you just mean that we should use a namedtuple to check that the input is valid (instead of staying `assert solar_distribution in ShadowCalculation.SOLAR_DISTRIBUTIONS`)?

honeybee_energy/properties/model.py
```diff
@@ -261,6 +262,33 @@ def program_types(self):
...
+                  '',
+                  '',
+                  solar_distribution)
+        comments = ('name',
+                    'north axis',
+                    'terrain',
+                    'loads convergence tolerance',
+                    'temperature convergence tolerance',
+                    'solar distribution')
+        return generate_idf_string('Building', values, comments)
```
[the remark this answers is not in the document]
chriswmackey: Thank you. Good to know and I agree it's acceptable!

honeybee_energy/schedule/day.py
```diff
@@ -425,9 +425,16 @@ def from_dict(cls, data):
 
         return cls(data['name'], data['values'], times, interpolate)
 
-    def to_idf(self):
-        """IDF string representation of ScheduleDay object."""
-        fields = [self.name, '']
+    def to_idf(self, schedule_type_limit=None):
+        """IDF string representation of ScheduleDay object.
```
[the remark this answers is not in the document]
chriswmackey: Yes, that's a good idea. We can display it within a `codeblock::shell`. I added an issue here:
https://github.com/ladybug-tools/honeybee-energy/issues/37

honeybee_energy/properties/model.py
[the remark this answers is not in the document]
mostaphaRoudsari: Not just for checking. I was thinking that we should create an object for `SolarDistribution` then use that for input. So instead of `FullInteriorAndExteriorWithReflections` you pass `SolarDistribution.FullInteriorAndExteriorWithReflections`.

Now that I think about it enumerate module is not supported in Python 2 so it will be a pain to handle it. Let's leave it as is.

honeybee_energy/simulationparameter.py
chriswmackey: It would work but it wouldn't look good in the resulting idf file. Essentially, there's be no blank line in between each object, making it look like one huge object.

honeybee_energy/writer.py
chriswmackey: What alternative are you suggesting?

honeybee_energy/writer.py
chriswmackey: I agree what you have there is much cleaner.

honeybee_energy/writer.py
chriswmackey: Whenever the parent object hosting the `Shade` is an `Aperture` and that `Aperture` has a parent `Face`. This really should be the case for any energy-simulate-able `Model` but might not be the case if someone passes a shade from an orphaned aperture into this method. So the use of `try/except` here is because the `except` case is truly exceptional.

honeybee_energy/writer.py
chriswmackey: I do eventually create a set (a few lines after this) but there's a very good performance reason why I use this method first.

Let's say I have a model of 1000 Faces and each one has a construciton that needs to be written into EnergyPlus. 500 of these Faces actually have the same "interior wall" construction instance assigned to them.  If I make a set of all 1000 construciton instances, it's going to check whether each of those constructions is equal to any others in the set. This is really computationally intensive given that each and every property of each and every material in the construction is checked for equality. If I simply check whether the object instance is already in the array, it is orders of magnitude faster (the performance tests I ran showed that it ran 1000 times faster). So that is why we use this method here.

honeybee_energy/writer.py
chriswmackey: I see people prefer different ways but I still don't see a better solution for this example.

Issue #56, opened 2019-11-15
"extract_all_from_idf_file" might throw error if IDF contains mutiple week rules


zha:
I was trying to execute `extract_all_from_idf_file` for the following IDF command but failed. It gives me the error "ScheduleDay objects can be assigned to a ScheduleRuleset only once."
```
ScheduleTypeLimits,
  avail_limit,                 !- Name
  0,                                      !- Lower Limit Value {BasedOnField A3}
  1,                                      !- Upper Limit Value {BasedOnField A3}
  Discrete,                               !- Numeric Type
  availability;                           !- Unit Type


Schedule:Day:Interval,
  off,                                   !- Name
  avail_limit,                 !- Schedule Type Limits Name
  No,                                     !- Interpolate to Timestep
  24:00,                                  !- Time 1 {hh:mm}
  0;                                      !- Value Until Time 1
Schedule:Day:Interval,
  on,                                   !- Name
  avail_limit,                 !- Schedule Type Limits Name
  No,                                     !- Interpolate to Timestep
  24:00,                                  !- Time 1 {hh:mm}
  1;                                      !- Value Until Time 1


Schedule:Week:Daily, 
Weeks_always_off,                       !- Name
off,                                    !- Sunday Schedule:Day Name
off,                                    !- Monday Schedule:Day Name
off,                                    !- Tuesday Schedule:Day Name
off,                                    !- Wednesday Schedule:Day Name
off,                                    !- Thursday Schedule:Day Name
off,                                    !- Friday Schedule:Day Name
off,                                    !- Saturday Schedule:Day Name
off,                                    !- Holiday Schedule:Day Name
off,                                    !- SummerDesignDay Schedule:Day Name
off,                                    !- WinterDesignDay Schedule:Day Name
off,                                    !- CustomDay1 Schedule:Day Name
off;                                    !- CustomDay2 Schedule:Day Name


Schedule:Week:Daily,
Week_cooling_switch_on,                        !- Name
off,                                     !- Sunday Schedule:Day Name
off,                                     !- Monday Schedule:Day Name
on,                                    !- Tuesday Schedule:Day Name
on,                                    !- Wednesday Schedule:Day Name
on,                                    !- Thursday Schedule:Day Name
on,                                    !- Friday Schedule:Day Name
on,                                    !- Saturday Schedule:Day Name
off,                                    !- Holiday Schedule:Day Name
off,                                    !- SummerDesignDay Schedule:Day Name
[...]

PR #57, opened 2019-11-19
fix(ruleset): Bug fix to ScheduleRuleset.extract_all_from_idf_file

chriswmackey:
This fixes the bug found by @zha and reported in this issue: https://github.com/ladybug-tools/honeybee-energy/issues/56

Specifically, it addresses cases where the same ScheduleDay is referenced by multiple ScheduleWeeks in an IDF file.  This doesn't map as cleanly to the ScheduleRuleset schema or, rather, it creates cases where someone can edit the ScheduleDay of one ScheduleRule and this inadvertently edits the ScheduleDay of another ScheduleRule. I realize that this "inadvertent editing" situation happens in the IDF world and no one minds but we also know that the IDF world is mostly concerned with storing schedules for individual Models and really isn't meant to manage a library of schedules as honeybee is. This is apparent in the fact that IDF makes it difficult to view year schedules as self-contained units that can be called from a library when these IDF year schedules depend on week schedules and day schedules that might be referenced elsewhere in a Model. This "self-containment" of year schedules is something we have been able to achieve with the honeybee schema and I would rather not give it up just to address this case.

So, in this light, the best way to handle this case seems to be to duplicate the cross-referenced ScheduleDay and give it a new unique name so that it does not conflict with the other ScheduleDay.  This is what I have done in this PR.

### DIFF
--- a/honeybee_energy/schedule/ruleset.py
+++ b/honeybee_energy/schedule/ruleset.py
@@ -19,6 +19,7 @@ from ladybug.datatype.generic import GenericType
 
 import re
 import os
+import uuid
 
 
 @lockable
@@ -1094,10 +1095,21 @@ class ScheduleRuleset(object):
         """Get a dictionary of ScheduleRule objects from Schedule:Week strings."""
         week_schedule_dict = {}
         week_designday_dict = {}
+        used_sch_days = []  # track which ScheduleDays have already been used
         for sch_str in week_idf_strings:
             sch_str = sch_str.strip()
             rules = ScheduleRule.extract_all_from_schedule_week(sch_str, day_sch_dict)
             if sch_str.startswith('Schedule:Week:Daily,'):
+                # check that ScheduleDays aren't referenced by multiple ScheduleWeeks
+                for sch_rule in rules:
+                    day_name = sch_rule.schedule_day.name
+                    if day_name not in used_sch_days:
+                        used_sch_days.append(day_name)
+                    else:  # cross-referenced ScheduleDay; duplicate and rename it
+                        sch_rule.schedule_day = sch_rule.schedule_day.duplicate()
+                        new_n = '{}_{}'.format(day_name, str(uuid.uuid4()).split('-')[1])
+                        sch_rule.schedule_day.name = new_n
+                # get the summer and winter design days
                 ep_strs = parse_idf_string(sch_str)
                 summer_dd = day_sch_dict[ep_strs[9]]
                 winter_dd = day_sch_dict[ep_strs[10]]
--- a/honeybee_energy/schedule/ruleset.py
+++ b/honeybee_energy/schedule/ruleset.py
@@ -19,7 +19,6 @@ from ladybug.datatype.generic import GenericType
 
 import re
 import os
-import uuid
 
 
 @lockable
@@ -132,7 +131,7 @@ class ScheduleRuleset(object):
     def summer_designday_schedule(self):
         """Get or set the DaySchedule that will be used for the summer design day.
 
-        Note that, if this property is None, the default_day_schedule is what is
+        Note that, if this property is None, the default_day_schedule is
         ultimately written into the IDF for the summer design day.
         """
         return self._summer_designday_schedule
@@ -150,7 +149,7 @@ class ScheduleRuleset(object):
     def winter_designday_schedule(self):
         """Get or set the DaySchedule that will be used for the winter design day.
 
-        Note that, if this property is None, the default_day_schedule is what is
+        Note that, if this property is None, the default_day_schedule is
         ultimately written into the IDF for the winter design day.
         """
         return self._winter_designday_schedule
@@ -168,12 +167,15 @@ class ScheduleRuleset(object):
     def day_schedules(self):
         """Get a list of all unique ScheduleDay objects used in this ScheduleRuleset."""
         day_scheds = [self.default_day_schedule]
-        if self._summer_designday_schedule is not None:
+        if self._summer_designday_schedule is not None and not \
+                self._instance_in_array(self._summer_designday_schedule, day_scheds):
             day_scheds.append(self._summer_designday_schedule)
-        if self._winter_designday_schedule is not None:
+        if self._winter_designday_schedule is not None and not \
+                self._instance_in_array(self._winter_designday_schedule, day_scheds):
             day_scheds.append(self._winter_designday_schedule)
         for rule in self.schedule_rules:
-            day_scheds.append(rule._schedule_day)
+            if not self._instance_in_array(rule._schedule_day, day_scheds):
+                day_scheds.append(rule._schedule_day)
         return day_scheds
 
     @property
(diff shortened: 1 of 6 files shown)


chriswmackey:
I'm trying to understand what you are saying here @mostaphaRoudsari .  Are you saying that users would prefer the flat, cross-reference-able schema of IDF schedule elements (days and weeks) instead of the nested schema of Honeybee schedule elements (which can sometimes require duplication in the translation from IDF as you see addressed in this PR)?

I just find the flat schema tough to work with for the same reasons why we found the honeybee-core geometry schema was tough to work with when originally had it flat (just like it is in IDF):

- Making visualizations of year schedules from JSON is now much easier.  If you only want to load a single schedule from a model, you just get the name of that schedule and load all of the schedule elements that are nested underneath it (no need to lookup schedule elements under another part of the schema or interact with the schedule elements you don’t care about).
- Manually adding/removing year Schedules from the schema is now much easier in that you simply delete the Schedule with all nested objects underneath it (or simply add a new object schema to the “schedules” key). There's no need to look up all of the other schedule elements in the deletion process and see if they are referenced by other year schedules.
- We don't have to worry about bi-directional name references within the year Schedule and its elements, which avoids confusion about which reference is used to re-establish a parent-child relationship.
- Not having to include all of the name references makes the schema lighter with fewer lines that people need to wade through to edit or understand it.

mostaphaRoudsari:
> Are you saying that users would prefer the flat, cross-reference-able schema of IDF schedule elements (days and weeks) instead of the nested schema of Honeybee schedule elements

no

> which can sometimes require duplication in the translation from IDF as you see addressed in this PR

The duplication is what I'm not sure about. Does it make sense to say if the names are referring to the exact same object then they should be allowed to be the same.

chriswmackey:
I guess another option that still lets us reap all of the benefits above while (maybe?) giving users "what they want" is to allow cross-referencing of individual day schedules within a single ScheduleYear/ScheduleRuleset. In this case, the schema for `ScheduleRuleset` would just have its own `schedule_days` key under which all of the relevant `ScheduleDays` sit. Then they Python object could work similarly by editing all of the schedule rules of the Ruleset that are using a given schedule day when you edit that schedule day. I could maybe see that as nice.  We just wouldn't allow the editing of all the ShceduleYears/ScheduleRulesets that use a given ScheduleDay (the way IDF does) because I know that will get confusing and it kills all of the benefits of the nested schema noted above.

I could make such a change but the question I am hung up on is whether this is something people actually want. Frankly, it's not something that I really want and I _feel_ like it would still be prone to "accidental editing".  Do you feel that it is something you want @mostaphaRoudsari ?  If so, we should probably make this change sooner than later since it really won't be possible to make such a large structural change without breaking things once the plugin is out in the wild.

chriswmackey:
Also, judging from your comment above, I think you might not be seeing that referencing by name (thereby allowing cross references) is mutually exclusive from a nested schema that does not use names to reference but rather uses the nested parent-child relationship to "reference".  Is it clear now what I'm talking about now?

chriswmackey:
I just wanted to add one other note, which is that this issue we are discussing here is the exact same as the one we discussed in our last call with Dan. In response to the duplication that we are discussing here, Dan's advice was:
> Duplication isn't necessarily a bad thing as long as it's useful.

From the perspective of creating and editing ScheduleRulesets and understanding that a ScheduleRuleset is just a single default ScheduleDay on top of which rules are applied, this duplication was definitely useful.  But I think we are realizing that, if serialization to/from IDF is going to be more common than creating/editing schedules as rule sets, then the utility of a clean serialization might outweigh the utility of duplication.

Let's give ourselves a little time to think about this and also see if we can understand the use case that led @zha to find this bug in the first place. I noticed that [the temporary solution that @zha implemented](https://github.com/zha/honeybee-energy/commit/c9150325cfb8df647394b67c417c5925516a40c3) is very much in line with the solution I just mentioned where we allow cross-referencing of individual day schedules within a single ScheduleRuleset. So that might be an indication that this type of solution is more useful.

chriswmackey:
After thinking about this for a few days, understanding @zha's use case, and drawing up [a list of pros/cons for the ScheduleDay duplication](https://docs.google.com/document/d/1ydgBbOzzYP8AeWdGCJaCqMXcDs8i0uMr3OyJN9RQRRw/edit?usp=sharing), I have come to realize that allowing cross-references of ScheduleDays within a single ScheduleRuleset should be permitted.  This will create a bit of a headache for us when we eventually make GUIs to edit schedule JSONs but it is a manageable headache that seems worthwhile for the cleaner serialization to/from IDF/OpenStudio models and the way it better handles certain edge cases.

While having the same ScheduleDay used multiple times within the same ScheduleRuleset should be allowed, I still plan to enforce that each ScheduleDay object have only one parent ScheduleRulest, which is not something that IDF or OSM enforces but this lack of enforcement creates a huge headache when trying to edit annual schedules in isolation of other schedules or when adding/removing annual schedules to/from models or schedule libraries.  I imagine that this headache is a primary reason why the OpenStudio GUI has never supported the editing of schedules (or, if it ever did, it was probably going to be very similar to the IDFEditor). The lack of enforcement also makes is easy to create illogical situations like having different ScheduleTypeLimits for two ScheduleDays used by the same ScheduleYear. Another crazy situation is having different heating and cooling design day schedules depending on which week of the year the design day falls under (this makes it almost impossible to track down what the actual design day schedule is unless you have the ddy with you).  So this is all stuff that I really want to avoid.

With this said, @zha's case has showed me that the IDF import should not fail no matter what cross-references there are since this is bound to be a common case. I also realize that renaming the ScheduleDay is going to be more annoying than helpful in these situations so I think I will automatically duplicate the schedule object upon import from IDF in this case but not change the name.  In the `model.to.idf` method, I will add some code to sense when ScheduleDays are equal one another and only write a given ScheduleDay once. We will need a similar check in the energy-model-measure. This way, serialization to/from IDF models is all supported with corss-references but ScheduleRulesets are still self contained units that can be edited in isolation of one another and in isolation of a Model.

chriswmackey:
Now that I decided to go in a different direction, this one needs your review again @mostaphaRoudsari .

honeybee_energy/schedule/ruleset.py
```diff
@@ -828,9 +841,19 @@ def extract_all_from_idf_file(idf_file):
                     rule.start_date = st_date
                     rule.end_date = end_date
                 all_rules.extend(rules)
-            default_day_schedule = all_rules[0].schedule_day
+            # gather the other day schedules
             summer_dd_sch, winter_dd_sch = week_dd_dict[year_sch[2]]
             schedule_type = sch_type_dict[year_sch[1]] if year_sch[1] != '' else None
+            # check to be sure the schedule day's don't already have a parent
```
mostaphaRoudsari:
```suggestion
            # check to be sure the schedule days don't already have a parent
```

honeybee_energy/schedule/ruleset.py
```diff
@@ -1169,6 +1197,20 @@ def _get_avg_week(name, schedules, weights, timestep_resolution, rule_indices):
             avg_mtx[5], avg_mtx[6], avg_mtx[7], timestep_resolution,
             schedules[0].schedule_type_limit, avg_mtx[8], avg_mtx[9])
 
+    @staticmethod
+    def _instance_in_array(object_instance, object_array):
+        """Check if a specific object instance is already in an array.
+
+        This can be much faster than  `if object_instance in object_arrary`
```
mostaphaRoudsari: I believe you have done your research on this. At the same time does this logic mean that the way `==` is implemented in the objects that check hasn't been implemented efficiently? I don't mean to go back and revise them all right now but just something to think about.

chriswmackey: It's not that the method is implemented inefficiently.  It's that schedules have a lot of properties to check in order to know that they are equal to one another.  If the object in question was something as simple as a material, all that we need to check in order to know if the materials are equal is the name and a handful of properties. So this is a fast check.  But for ScheduleRulesets, there are often several ScheduleDays to check and we need to check that all of these ScheduleDay values and times are equal to one another. There's also ScheduleTypelimits and other things to check like rules. So it takes significantly longer than just checking if something is the same instance.  Also, with these changes I made in this PR, the intention is that  someone uses the same ScheduleDay instance when they have a cross-referenced ScheduleDay (not two different ScheduleDay instances that are equivalent to one another). So this method here actually enforces this intention and helps discourage the latter.

PR #64, opened 2019-12-19
feat(config): Add a config module to handle folder locations

chriswmackey:
The config file handles several commonly-accessed folders including the folders for openstudio/ energyplus installations, the user-defined library of templates (eg. constructions, schedules, etc.), the default simulation folder, and the location of the energy_model_measure.

### DIFF
--- a/honeybee_energy/config.py
+++ b/honeybee_energy/config.py
@@ -1,0 +1,394 @@
+"""Honeybee_energy configurations.
+
+Import this into every module where access configurations are needed.
+
+Usage:
+    from honeybee_energy.config import folders
+    print(folders.energyplus_path)
+    print(folders.openstudio_path)
+    folders.energyplus_path = "C:/EnergyPlusV9-0-1"
+"""
+import honeybee.config as hb_config
+
+import os
+import sys
+import json
+
+
+class Folders(object):
+    """Honeybee_energy folders.
+
+    Args:
+        config_file: The path to the config.json file from which folders are loaded.
+            If None, the config.json module included in this package will be used.
+            Default: None.
+        mute: If False, the paths to the various folders will be printed as they
+            are found. If True, no printing will occur upon initialization of this
+            class. Default: True.
+
+    Properties:
+        * openstudio_path
+        * openstudio_exe
+        * energyplus_path
+        * energyplus_exe
+        * energy_model_measure_path
+        * template_library_folder
+        * construction_lib
+        * constructionset_lib
+        * schedule_lib
+        * programtype_lib
+        * config_file
+        * mute
+    """
+
+    def __init__(self, config_file=None, mute=True):
+        # set the mute value
+        self.mute = bool(mute)
+
+        # load paths from the config JSON file 
+        self.config_file  = config_file
+
+    @property
+    def openstudio_path(self):
+        """Get or set the path to OpenStudio installation folder.
+        
+        This is the "bin" directory for OpenStudio installation (the one that
+        contains the openstudio executable file).
+        """
+        return self._openstudio_path
+
+    @openstudio_path.setter
+    def openstudio_path(self, path):
+        exe_name = 'openstudio.exe' if os.name == 'nt' else 'openstudio'
+        if not path:  # check the PATH and then the default installation location
+            path, os_exe_file = self._which(exe_name)
+            if path is None:  # search within the default installation location
+                path, os_exe_file = self._find_openstudio_folder()
+        else:
+            os_exe_file = os.path.join(path, exe_name)
+
+        if path:  # check that the OpenStudio executable exists in the path
+            assert os.path.isfile(os_exe_file), \
+                '{} is not a valid path to an openstudio installation.'.format(path)
+
+        #set the openstudio_path
+        self._openstudio_path = path
+        self._openstudio_exe = os_exe_file
+        if path and not self.mute:
+            print("Path to OpenStudio is set to: %s" % path)
+    
+    @property
+    def openstudio_exe(self):
+        """Get the path to the executable openstudio file."""
+        return self._openstudio_exe
+
+    @property
+    def energyplus_path(self):
+        """Get or set the path to EnergyPlus installation folder."""
+        return self._energyplus_path
+    
+    @energyplus_path.setter
+    def energyplus_path(self, path):
+        exe_name = 'energyplus.exe' if os.name == 'nt' else 'energyplus'
+        if not path:  # check the PATH and then the default installation location
+            path, ep_exe_file = self._which(exe_name)
+            if path is None:  # search within the default installation location
+                path, ep_exe_file = self._find_energyplus_folder()
+        else:
+            ep_exe_file = os.path.join(path, exe_name)
+        
+        if path:  # check that the Energyplus executable exists in the installation
+            assert os.path.isfile(ep_exe_file), \
+                '{} is not a valid path to an energyplus installation.'.format(path)
+
+        # set the energyplus_path
+        self._energyplus_path = path
+        self._energyplus_exe = ep_exe_file
+        if path and not self.mute:
+            print("Path to EnergyPlus is set to: %s" % self._energyplus_path)
+
+    @property
+    def energyplus_exe(self):
+        """Get the path to the executable energyplus file."""
+        return self._energyplus_exe
+    
+    @property
+    def energy_model_measure_path(self):
+        """Get or set the path to the energy_model_measure translating to OpenStudio.
+        
+        This folder must have the following sub-folders in order to be valid:
+            * ladybug - ruby library with modules for model translation to OpenStudio.
+            * measures - folder with the actual measures that run the translation.
+        """
+        return self._energy_model_measure_path
+    
+    @energy_model_measure_path.setter
+    def energy_model_measure_path(self, path):
+        if not path:  # check the default locations of the energy_model_measure
+            path = self._find_energy_model_measure_path()
+
+        # check that the library's sub-folders exist
+        if path:
+            assert os.path.isdir(os.path.join(path, 'ladybug')), \
+                '{} lacks a "ladybug" folder for the translation ruby library.'.format(path)
+            assert os.path.isdir(os.path.join(path, 'measures')), \
+                '{} lacks a "measures" folder.'.format(path)
+
+        # set the energy_model_measure_path
+        self._energy_model_measure_path = path
+        if path and not self.mute:
+            print('Path to the energy_model_measure is set to: '
+                    '{}'.format(self._energy_model_measure_path))
+    
+    @property
+    def template_library_folder(self):
+        """Get or set the path to the library of templates loaded to honeybee_energy.lib.
+        
+        This folder must have the following sub-folders in order to be valid:
+            * constructions - folder with IDF files for materials + constructions.
+            * constructionsets - folder with JSON files of abridged ConstructionSets.
+            * schedules - folder with IDF files for schedules.
+            * programtypes - folder with JSON files of abridged ProgramTypes.
+        """
+        return self._template_library_folder
+    
+    @template_library_folder.setter
+    def template_library_folder(self, path):
+        if not path:  # check the default locations of the template library
+            path = self._find_template_library_folder()
+        
+        # gather all of the sub folders underneath the master folder
+        self._construction_lib = os.path.join(path, 'constructions') if path else None
+        self._constructionset_lib = os.path.join(path, 'constructionsets') if path else None
+        self._schedule_lib = os.path.join(path, 'schedules') if path else None
+        self._programtype_lib = os.path.join(path, 'programtypes') if path else None
+
+        # check that the library's sub-folders exist
+        if path:
+            assert os.path.isdir(self._construction_lib), \
+                '{} lacks a "constructions" folder.'.format(path)
+            assert os.path.isdir(self._constructionset_lib), \
+                '{} lacks a "constructionsets" folder.'.format(path)
+            assert os.path.isdir(self._schedule_lib), \
+                '{} lacks a "schedules" folder.'.format(path)
+            assert os.path.isdir(self._programtype_lib), \
+                '{} lacks a "programtypes" folder.'.format(path)
+
+        # set the template_library_folder
+        self._template_library_folder = path
+        if path and not self.mute:
+            print('Path to the template_library_folder is set to: '
+                    '{}'.format(self._template_library_folder))
+    
+    @property
+    def construction_lib(self):
+        """Get the path to the construction library in the template_library_folder."""
+        return self._construction_lib
+    
+    @property
+    def constructionset_lib(self):
+        """Get the path to the constructionset library in the template_library_folder."""
+        return self._constructionset_lib
+    
+    @property
+    def schedule_lib(self):
+        """Get the path to the schedule library in the template_library_folder."""
+        return self._schedule_lib
+    
+    @property
+    def programtype_lib(self):
+        """Get the path to the programtype library in the template_library_folder."""
+        return self._programtype_lib
+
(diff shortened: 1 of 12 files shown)


honeybee_energy/config.py
```diff
@@ -0,0 +1,322 @@
...
+        folders = Folders(mute=False)
+        print(folders.energyplus_path)
+    """
+
+    def __init__(self, mute=True):
+        # set the mute value
+        self.mute = bool(mute)
+
+        # load paths from the config JSON file within this package
+        config_file = os.path.join(os.path.dirname(__file__), 'config.json')
```
mostaphaRoudsari: I know this is how config.json is implemented in honeybee but there are a couple of issues with this implementation. The most important one is that the user values will be overwritten every time that the library is updated.

I think at the minimum we should write config_file as a property so one can change it to a different file. This will be useful for our plugins in-which we probably want to create the file in the installation folder and not inside the Python libraries installation folder.

Something else to think about is if we want to have a config file for honeybee and each extension (e.g. honeybee-energy) extends the config file? With that approach folders will be modified from a single place. We can also add a component to the interface to set these values for each extension.

honeybee_energy/config.py
```diff
@@ -0,0 +1,322 @@
...
+        This is the "master" directory for OpenStudio installation (the one that
+        includes EnergyPlus, Ruby, etc.). It must contain a "bin" folder
+        with the openstudio executable in order to be valid.
+        """
+        return self._openstudio_path
+
+    @openstudio_path.setter
+    def openstudio_path(self, path):
+        if not path:  # check the default installation folders
+            path = self._find_openstudio_folder()
```
mostaphaRoudsari: I think we should remove all the find methods and make it explicit as part of the process by users to set the folders.

A better approach is most likely using environment variables and stop looking into path for folders.

honeybee_energy/config.py
```diff
@@ -0,0 +1,322 @@
...
+
+    @property
+    def python_package_path(self):
+        """Get the path to where this Python package is installed."""
+        return os.path.split(os.path.dirname(__file__))[0]
+
+    @property
+    def python_exe_path(self):
+        """Get the path to the Python.exe file."""
+        return sys.executable
```
mostaphaRoudsari: This should be part of honeybee-core.

honeybee_energy/config.py
```diff
@@ -0,0 +1,322 @@
...
+        
+        Args:
+            file_path: Path to a JSON file containing the file paths. A sample of this
+                JSON is the config.json file within this library.
+        """
+        # check the default file path
+        assert os.path.isfile(str(file_path)), \
+            ValueError('No such a file as {}'.format(file_path))
+
+        # set the default paths to be all blank / defauktes
```
mostaphaRoudsari:
```suggestion
        # set the default paths to be all blank / defaults
```

honeybee_energy/config.py
```diff
@@ -0,0 +1,322 @@
...
+        default_path = {
+            "energyplus_path": r'',
+            "openstudio_path": r'',
+            "template_library_folder": r'',
+            "default_simulation_folder": r'',
+            "energy_model_measure_path": r''
+        }
+
+        with open(file_path, 'rb') as cfg:
```
mostaphaRoudsari: `rb` can cause issues in Python 2 versus 3.

```suggestion
        with open(file_path, 'r') as cfg:
```

honeybee_energy/config.py
```diff
@@ -0,0 +1,322 @@
...
+            path = cfg.read()
+            try:
+                paths = json.loads(path)
+            except Exception as e:
+                print('Failed to load paths from {}.\n{}'.format(file_path, e))
```
mostaphaRoudsari: We should probably just use `json.load(cfg)`?

honeybee_energy/config.py
```diff
@@ -0,0 +1,322 @@
...
+        
+        This is usually next to the Python pacakges when it's installed for a plugin.
+        """
+        measure_path = os.path.join(self.python_package_path, 'energy_model_measure')
+        if not os.path.isdir(measure_path):
+            return  # No energy_model_measure is installed
+        return measure_path
+
+    @staticmethod
+    def _find_openstudio_folder():
```
mostaphaRoudsari: This is part of a larger conversation but I think all the execution related information like this one should be moved outside the libraries. The only case that I can imagine keeping them here is useful is for finding C# libraries if we decide to use them in the code.

I understand with the current design we will need them for executing the simulations but as I commented before I think we should follow a PATH based approach instead of finding the installation folder. This will also help to ensure a batch / bash file will run successfully on any machine regardless of the operating system and installation folder of the computer that is used to create it.

tests/config_test.py
```diff
@@ -0,0 +1,31 @@
+# coding=utf-8
+from honeybee_energy.config import folders
+
+import pytest
+
+
+def test_config_init():
+    """Test the initialization of the config module and basic properties."""
+    assert hasattr(folders, 'energyplus_path')
+    assert folders.energyplus_path is None or isinstance(folders.energyplus_path, str)
```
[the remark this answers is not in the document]
chriswmackey: "the real issue" is not very portable, though. I still want the tests to pass if I'm running the tests on a machine with or without EnergyPlus.

honeybee_energy/lib/_loadtypelimits.py
```diff
@@ -9,8 +10,7 @@
 
 
 # load schedule types from the default and user-supplied files
-cur_dir = os.path.dirname(__file__)
-schedule_lib = os.path.join(cur_dir, 'library', 'schedules')
+schedule_lib = os.path.join(folders.template_library_folder, 'schedules')
```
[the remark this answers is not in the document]
chriswmackey: I originally have 5 separate attributes for : materials, constructions, constructionists, schedules and programtypes but this was difficult to use. I can add such properties as read-only derived from the template_library variable if that is what you mean.

honeybee_energy/writer.py
```diff
@@ -387,9 +387,14 @@ def model_to_idf(model, schedule_directory=None,
                 sched_strs.extend([year_schedule] + week_schedules + day_scheds)
         except AttributeError:  # ScheduleFixedInterval
             if sched_dir is None:
-                sched_dir = schedule_directory if schedule_directory is not None \
-                    else os.path.join(os.environ['USERPROFILE'], 'honeybee',
-                                      'unnamed', 'schedules')
+                if schedule_directory is None:
+                    try:  # windows
```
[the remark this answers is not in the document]
chriswmackey: Ok. I can move this to honeybee_core.

honeybee_energy/lib/_loadschedules.py
```diff
@@ -9,8 +10,7 @@
 
 
 # load schedules from the default and user-supplied files
-cur_dir = os.path.dirname(__file__)
-schedule_lib = os.path.join(cur_dir, 'library', 'schedules')
+schedule_lib = os.path.join(folders.template_library_folder, 'schedules')
```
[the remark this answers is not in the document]
chriswmackey: I agree with the first point that it would be good to not rebuild it in different module each time. So I will definitely add a read-only property. I'm not so sold on the second item, though, since I think it will be a more common case to reset the folder for all of the libraries at once (materials, schedules, programtypes, constructionists) rather than setting individual libraries. But having everything use the same properties leaves this option open down the line if we find that we need it.

honeybee_energy/lib/_loadtypelimits.py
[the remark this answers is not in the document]
mostaphaRoudsari: > I can add such properties as read-only derived from the template_library variable if that is what you mean.

Yes.

tests/config_test.py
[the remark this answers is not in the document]
mostaphaRoudsari: I'm still not sure what are you trying to test here. What does the value being None or string tells us?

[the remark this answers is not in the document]
chriswmackey: It's running the code to make sure that it doesn't fail. If there's no installation found, it will be None. Otherwise it will be a string to the path.

honeybee_energy/config.py
chriswmackey: Makes sense. Will do.

honeybee_energy/config.py
chriswmackey: I agree that a PATH-based solution would be nice and it's definitely what we should do for Radiance but I don't think that a purely PATH-based approach is going to work well for EnergyPlus/OpenStudio because the installers for EnergyPlus/OpenStudio do not include any option to add the executables to the PATH. This means that we either have to have people manually add it to the PATH, which I know isn't realistic for the majority of users attending our workshops, or we try to add it to the path using a script, which is bound to run into permissions issues on some machines unless this script is a part of a signed installer.
So, in the long run, I think a purist PATH-based approach is going to mean it works on fewer machines rather than more machines. But I would definitely be in favor of  an approach that tries to check the PATH first and then, if it doesn't find an installation there, it looks in the usual default location. The PATH-based approach will be useful on our cloud service or for those users who might buy our installer but I think we're easily going to loose the majority of the users not going this route unless we include a backup way to find the installation.
Let me know if you agree and I will implement the dual solution.

honeybee_energy/config.py
chriswmackey: I was thinking about this case when I put together this class and you can see that the publicly-available `load_from_file` method in this PR addresses this exact concern. We could also do this as a property that sets all of the other properties of the class but, at that point, it feels like it really should be a method as I have now.

And I hear you that there may be cases where you don't want to overwrite the config file but, at the least, I think you need a config.json file in the python package to catch the case that we don't find a config.json file in the user folder.

mostaphaRoudsari: That's a good option to have but `load_from_file` ideally should be part of the setter. Right now one might set the config file without calling load_from_file and that won't trigger loading the files.

``` python
@property 
def config_file(self):
    return self._config_file

@config_file.setter
def config_file(self, cfg):
    # extra checks
    self._load_from_file(cfg)
    # or something similar to this
```

honeybee_energy/config.py
mostaphaRoudsari: That can work.

There will not be permission issues. You can pass the PATH when starting a subprocess. User doesn't need to know about it.

tests/config_test.py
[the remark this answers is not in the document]
chriswmackey: I realize that this is not the best example of a test but running the code to make sure there aren't any typos in it still has value.

honeybee_energy/lib/_loadconstructionsets.py
```diff
@@ -15,8 +16,7 @@
 
 
 # load construction sets from the default and user-supplied files
-cur_dir = os.path.dirname(__file__)
-constr_lib = os.path.join(cur_dir, 'library', 'constructionsets')
+constr_lib = os.path.join(folders.template_library_folder, 'constructionsets')
```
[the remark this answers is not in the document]
chriswmackey: You mean `constructionset_lib` but, yes, I agree.

honeybee_energy/config.py
chriswmackey: I see. So you think there are cases where we will want to check the location of the `config_file` being used. Ok. I'll make it a set-able property and make `load_from_file` hidden.

honeybee_energy/config.py
chriswmackey: Ok.

honeybee_energy/config.py
chriswmackey: Done!

honeybee_energy/config.py
chriswmackey: I agree that there won't be permission issues when the user is an admin or they are using an executable installer that they had to get admin rights to run. But, if there were no permissions issues in any case, you wouldn't have people asking [questions like this](https://serverfault.com/questions/33681/how-can-i-modify-a-user-s-path-environment-variable-without-logging-out) or there wouldn't be a little shield icon next to the menu that you use to edit the PATH manually on Windows:
image

chriswmackey: More proof that you need admin privileges to edit the PATH:
http://www.kscodes.com/misc/how-to-set-path-in-windows-without-admin-rights/

mostaphaRoudsari: You are confusing system variables and environmental variables for a subprocess. Those two are different. You can pass the PATH when **starting a subprocess** using the `env` argument. See the documentation here: https://docs.python.org/2/library/subprocess.html#popen-constructor

This is similar to how we set-up the PATH in every bat file. That does not need admin privileges.

honeybee_energy/writer.py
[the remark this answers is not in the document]
chriswmackey: Done. The new PR for honeybee_core is here:
https://github.com/ladybug-tools/honeybee-core/pull/46

honeybee_energy/config.py
chriswmackey: Done!

honeybee_energy/config.py
chriswmackey: Done.

tests/config_test.py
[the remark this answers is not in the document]
mostaphaRoudsari: > the code to make sure there aren't any typos in it still has value

I agree with this part of the test: `assert hasattr(folders, 'energyplus_path')`. My comment was on the second part.

honeybee_energy/lib/_loadconstructions.py
```diff
@@ -13,8 +14,7 @@
 
 
 # load materials and constructions from the default and user-supplied files
-cur_dir = os.path.dirname(__file__)
-construction_lib = os.path.join(cur_dir, 'library', 'constructions')
+construction_lib = os.path.join(folders.template_library_folder, 'constructions')
```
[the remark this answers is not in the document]
chriswmackey: Done

honeybee_energy/lib/_loadconstructionsets.py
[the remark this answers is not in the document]
chriswmackey: Done

honeybee_energy/lib/_loadtypelimits.py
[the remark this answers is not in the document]
chriswmackey: Ok done.

honeybee_energy/lib/_loadschedules.py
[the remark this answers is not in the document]
chriswmackey: I added the read-only property.

honeybee_energy/lib/_loadprogramtypes.py
```diff
@@ -12,8 +13,7 @@
 
 
 # load program types from the default and user-supplied files
-cur_dir = os.path.dirname(__file__)
-program_lib = os.path.join(cur_dir, 'library', 'programtypes')
+program_lib = os.path.join(folders.template_library_folder, 'programtypes')
```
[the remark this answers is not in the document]
chriswmackey: Done

honeybee_energy/config.py
chriswmackey: I'm going to resolve this one since we're already discussing this in your other comments.

honeybee_energy/config.py
chriswmackey: Ok, you will see that I implemented it in a way that, if there's no specification of the installation path in the config.json, it will check to see if the executable file exists in the PATH. Then, if it doesn't find it there, it will search in the default installation location and attempt to use the most recent version of EnergyPlus/OpenStudio it finds.

This means that we can eventually do some stuff with the PATH down the line when we figure out how we want to get around the admin access issue. But it will work now, even for people who don't have admin access on their machine.

chriswmackey:
I think I addressed all of your comments @mostapharoudsari . It's ready for your review again.

PR #65, opened 2019-12-26
feat(run): Add methods to run EnergyPlus/OpenStudio on Unix

chriswmackey:
I got the chance to borrow Rosi's Macbook for a few hours and it was enough for me to develop and test some functions that could run EnergyPlus successfully on Mac. This commit also includes a bug fix to the previous code that I had implemented to find the EnergyPlus/OpenStudio installation on Mac.

### DIFF
--- a/honeybee_energy/run.py
+++ b/honeybee_energy/run.py
@@ -8,95 +8,11 @@ from __future__ import division
 
 import os
 import json
+import subprocess
 
 from .config import folders
 
-from ladybug.futil import write_to_file_by_name, copy_files_to_folder, copy_file_tree
-
-
-def run_idf_windows(idf_file_path, epw_file_path):
-    """Run an IDF file through energyplus on a Windows-based operating system.
-
-    A batch file will be used to run the simulation.
-
-    Args:
-        idf_file_path: The full path to an IDF file.
-        epw_file_path: The full path to an EPW file.
-
-    Returns:
-        sql -- Path to a .sqlite file containing all simulation results.
-            Will be None if no file exists.
-        eio -- Path to a .eio file containing properties of the model, including
-            the size of HVAC objects. Will be None if no file exists.
-        rdd -- Path to a .rdd file containing all possible outputs that can be
-            requested from the simulation. Will be None if no file exists.
-        html -- Path to a .html file containing all summary reports.
-            Will be None if no file exists.
-    """
-    # check and prepare the input files
-    directory = prepare_idf_for_simulation(idf_file_path, epw_file_path)
-
-    # write a batch file
-    expand_path = os.path.join(folders.energyplus_path, 'ExpandObjects')
-    run_path = os.path.join(folders.energyplus_path, 'EnergyPlus')
-    working_drive = directory[:2]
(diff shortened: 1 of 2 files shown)


honeybee_energy/run.py
```diff
@@ -36,24 +42,83 @@ def run_idf_windows(idf_file_path, epw_file_path):
...
     # run the batch file
-    os.system(os.path.join(directory, 'in.bat'))
+    os.system(batch_file)
 
     # output the simulation files
-    return _output_files_from_directory(directory)
+    return _output_energyplus_files(directory)
+
+
+def run_idf_unix(idf_file_path, epw_file_path, expand_objects=True):
```
mostaphaRoudsari: Since these methods run the simulation I think we should have a single run_idf which senses the platform and does it correctly based on the operating system.

honeybee_energy/run.py
```diff
@@ -69,34 +134,63 @@ def run_osw_windows(osw_json, measures_only=True):
...
+
+    # Write the shell script to call OpenStudio CLI
+    measure_str = '-m ' if measures_only else ''
+    shell = '#!/usr/bin/env bash\n\ncd {}\n"openstudio" run {}-w {}'.format(
+        folders.openstudio_path, measure_str, osw_json)
+    shell_file = os.path.join(directory, 'run_workflow.sh')
+    write_to_file(shell_file, shell, True)
+
+    # make the shell script executable using subprocess.check_call
+    # this is more reliable than native Python chmod on Mac
```
mostaphaRoudsari: I don't know what you mean here by native Python chmod but the code below works just fine on Mac:

```python
def change_permission(exec_file):
    st = os.stat(exec_file)
    os.chmod(exec_file, st.st_mode | 0o111)
```

honeybee_energy/run.py
```diff
@@ -215,7 +313,31 @@ def prepare_idf_for_simulation(idf_file_path, epw_file_path):
     return directory
 
 
-def _output_files_from_directory(directory):
+def _prepare_osw(osw_json):
```
mostaphaRoudsari: This is more like checking for osw. prepare_osw sounds like like a method that you prepare the file itself.

honeybee_energy/run.py
chriswmackey: I was trying to get your suggestion to work on Rosi's Mac. I tried several variations of it but it just was not working when it was used from Rhino 6's IronPython. Maybe it will work with cPython but I am positive that what I have in place now works from Rhino 6's IronPython on Mac.

honeybee_energy/run.py
chriswmackey: Good suggestion. I will implement this.

honeybee_energy/run.py
chriswmackey: Ok. Will  change it to `check_osw`.

honeybee_energy/run.py
chriswmackey: Done.

honeybee_energy/run.py
chriswmackey: Done.

PR #67, opened 2020-01-07
feat(sizing): Add design_days to the SizingParameter object

chriswmackey:
Since these ultimately end up in the IDF and OSM, I think it makes sense to include them within the Python objects here rather than have users upload a .ddy file that might often have more design days than they actually need.

### DIFF
--- a/honeybee_energy/simulation/sizing.py
+++ b/honeybee_energy/simulation/sizing.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-"""Global sizing parameters dictating scaling factors for all zone peak loads."""
+"""Parameters with criteria for sizing the heating and cooling system."""
 from __future__ import division
 
 from ..reader import parse_idf_string
@@ -7,20 +7,27 @@ from ..writer import generate_idf_string
 
 from honeybee.typing import float_positive
 
+from ladybug.ddy import DDY
+from ladybug.designday import DesignDay
+from ladybug.location import Location
+
 
 class SizingParameter(object):
-    """Global sizing parameters dictating scaling factors for all zone peak loads.
+    """Sizing parameters with criteria for sizing the heating and cooling system.
 
     Properties:
+        * design_days
         * heating_factor
         * cooling_factor
     """
-    __slots__ = ('_heating_factor', '_cooling_factor')
+    __slots__ = ('_design_days', '_heating_factor', '_cooling_factor')
 
-    def __init__(self, heating_factor=1.25, cooling_factor=1.15):
+    def __init__(self, design_days=None, heating_factor=1.25, cooling_factor=1.15):
         """Initialize SizingParameter.
 
         Args:
+            design_days: An array of Ladybug DesignDay objects that represent the
+                criteria for which the HVAC systems will be sized. Default: None.
             heating_factor: A number that will get multiplied by the peak heating load
                 for each zone in the model in order to size the heating system for
                 the model. Must be greater than 0. Default: 1.25.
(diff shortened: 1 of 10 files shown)


honeybee_energy/simulation/sizing.py
```diff
@@ -7,30 +7,59 @@
...
         self.heating_factor = heating_factor
         self.cooling_factor = cooling_factor
 
+    @property
+    def design_days(self):
+        """Get or set an array of Ladybug DesignDay objects for sizing criteria."""
+        return tuple(self._design_days)
+
+    @design_days.setter
+    def design_days(self, value):
```
mostaphaRoudsari: I personally find having the combination of setter and `add_*` methods confusing. I think you can just have a range of `add_*` methods that one can use to add design days after initiating the object.

Also it might makes sense to have methods to remove_design_days or clear_design_days?

honeybee_energy/simulationparameter.py
```diff
@@ -159,11 +159,8 @@ def global_geometry_rules(self):
              Counterclockwise,         !- vertex entry direction
              Relative;                 !- coordinate system
         """
-        values = ('UpperLeftCorner',
-                  'Counterclockwise',
-                  'Relative')
-        comments = ('starting vertex position',
-                    'vertex entry direction',
+        values = ('UpperLeftCorner', 'Counterclockwise', 'Relative')
```
mostaphaRoudsari: This comment is not related to this line but the name of the module. Should we start break down the name of modules with long names. For instance call this one `simulation_parameter`? I know it is not best practice but it is valid and it is much easier to read.

This is one of those topics that we should discuss before the public release since it will create breaking changes.

chriswmackey: So there is a `simulation` module within honeybee and so I could just put this file within that module and call it `parameter`. So you access it by typing `honeybee.simulation.parameter`. But this SimulationParameter class is the parent class to all of the other sub-modules and classes in the `simulation` module. So I moved it out of the module to distinguish it. Do you think this was a mistake or do you know of any best practice here?  I could really go either way on it so I'll implement whatever you think is best.

honeybee_energy/simulation/sizing.py
chriswmackey: I'll add methods to remove design days.

The code of the setter still needs to be there, whether it's in `__init__` or it's in this setter here. So, whether it's public or private really isn't changing much in my view.

Do you not agree with the philosophy of "whatever can be public should be public"?  Personally, I have just found myself in many more situations wishing something was exposed rather than wishing it was hidden.

honeybee_energy/simulationparameter.py
chriswmackey: I'm realizing that I probably should put it in the simulation module, even if there's no way to distinguish it as the parent object when it's in there. Let me know if you agree.

chriswmackey: Ok. I implemented the name change of `simulation.parameter`.

For my thoughts on the underscores in the module names, I think your talking about a monumental effort across the libraries that will break almost everything (nearly all of the libraries I have worked on have several of these combined-name modules). If we're going to do this, it's better to do it now rather than later when we'd have to put out a bunch of deprecation warnings about it.  Personally, I don't think the change is worth the week or so it would take me to follow it through all of the libraries and Grasshopper plugins. And, as you say, it's not even considered the best practice. So I'm not very inclined to change it.

honeybee_energy/simulation/sizing.py
mostaphaRoudsari: My only concern is that a setter can overwrite all the other values in the list in this case. Generally speaking I agree with having what can be public, public.

honeybee_energy/simulationparameter.py
mostaphaRoudsari: This is not what I was thinking but it doesn't look bad to me.

You are right that using underscore will break everything. If we think it is a good idea with might want to do it closer to the release when everything else is in place but would you be fine doing it for the newer modules which we might add from now on?

chriswmackey: I think we should have a longer conversation about this. I'd much rather keep things consistent  for now and I'd definitely want to understand your reasoning before I sink a lot of time into addressing this.

honeybee_energy/simulation/sizing.py
chriswmackey: I think that's just the nature of working with a powerful API that you can overwrite things. As they say, "with great power comes great responsibility."

PR #69, opened 2020-01-13
feat(hvac): Refactor HVAC to be extensible for future systems

chriswmackey:
This commit overhauls how HVAC systems are currently assigned to the Python objects and how they are represented in the schema.

Specifically, it makes it possible in the future to support HVAC systems that are assigned to multiple Rooms (like most of the HVAC  system templates in honeybee_legacy).  It also restructures the representation of HVAC within the Model dictionary schema in order to support this.

This commit also creates a base _HVACSystem class from which all HVAC systems will inherit and lays out simple ground rules for adding new HVAC classes to honeybee_energy. These ground rules will be followed when we eventually add the honeybee_energy_hvac extension.

Lastly, I added a few new features to the IdealAirSystem class. Notably, I added heating/cooling availability schedules both as a means to test out how we will collect schedules across future detailed HVAC systems and because @zha specifically requested them :)


honeybee_energy/writer.py
```diff
@@ -328,39 +326,32 @@ def model_to_idf(model, schedule_directory=None,
...
         from honeybee_energy.lib.programtypes import office_program
         from honeybee_energy.idealair import IdealAirSystem
-        from honeybee_energy.simulationparameter import SimulationParameter
+        from honeybee_energy.simulation.parameter import SimulationParameter
 
         # Get input Model
         room = Room.from_box('Tiny House Zone', 5, 10, 3)
         room.properties.energy.program_type = office_program
-        room.properties.energy.hvac = IdealAirSystem()
+        room.properties.energy.add_default_ideal_air()
```
mostaphaRoudsari: Are you thinking to add this for all the other systems too? I think the normal assignment is the better way to go than add a method for default ideal air.

tests/model_extend_test.py
```diff
@@ -381,11 +380,11 @@ def test_to_dict_single_zone():
...
+        room.properties.energy.hvac.name
 
     """
-    f_dir = 'C:/Users/chris/Documents/GitHub/energy-model-schema/app/models/samples/json'
-    dest_file = f_dir + '/model_single_zone_office.json'
+    f_dir = 'C:/Users/chris/Documents/GitHub/honeybee-schema/honeybee_schema/samples'
+    dest_file = f_dir + '/model_complete_single_zone_office.json'
     with open(dest_file, 'w') as fp:
         json.dump(model_dict, fp, indent=4)
     """
```
mostaphaRoudsari: I think we discussed this before. If you need code snippet like this put it in a different script and not test. If it needs to be updated on each release then it should be part of travis. Having it inside test files is not the best idea.

honeybee_energy/hvac/_base.py
```diff
@@ -0,0 +1,129 @@
...
+class _HVACEnumeration(object):
+    """Enumerates all installed HVAC classes."""
+    _HVAC_TYPES = {}
+
+    def __init__(self, import_modules=True):
+        if import_modules:
+            self._import_modules()
+
+        for clss in _HVACSystem.__subclasses__():
+            self._HVAC_TYPES[clss.__name__] = clss
```
mostaphaRoudsari: My first impression is that this is too much magic for creating a dictionary of HVAC system objects but you probably have a good reason for doing so.

honeybee_energy/hvac/_base.py
```diff
@@ -0,0 +1,129 @@
...
+
+from ..schedule.ruleset import ScheduleRuleset
+from ..schedule.fixedinterval import ScheduleFixedInterval
+
+from honeybee._lockable import lockable
+from honeybee.typing import valid_ep_string
+
+from os.path import dirname, basename, isfile, join
+from os import listdir
+from importlib import import_module
```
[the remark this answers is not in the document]
chriswmackey: So what do you recommend, then?  I guess that I can remove the ability to extend the `hvac` subpackage by dropping hvac modules into the folder since I realized that the way that I was planning to extend the subpackage with the future `honeybee-energy-hvac` repo isn't necessarily reliant on this.  In this case, I will hard code the fact that `IdealAirSystem` is the only HVAC system that comes with `honeybee_energy`. Would you be in support of this?

On another note, [the other place where we use this strategy is over in `ladybug.datatype`](https://github.com/ladybug-tools/ladybug/blob/master/ladybug/datatype/base.py#L7). In that case, though, we are really reliant on it in order to make all of the datatype modules nice and small and make the whole subpackage easily extensible with new data types. Please open an issue over there if you have a recommended alternative.

honeybee_energy/writer.py
chriswmackey: No. For all other systems, I'm expecting the user to create their specific HVAC system class instance like so:
```
import TheHVACSystemClass

the_hvac_system_class_instance = TheHVACSystemClass(name='My HVAC')
room.properties.energy.hvac = the_hvac_system_class_instance
```
I just wanted to have a shortcut for the `IdealAirSystem` because they are used so much all over the place and the convenience of one line of code over 3 is pretty helpful.

tests/model_extend_test.py
chriswmackey: I was just trying to avoid duplicating the ~30 lines of code before this that it takes to create each model. Now that things are settling down and the JSON samples generated with this code aren't likely to change as much, I'll send a PR after this one that duplicates these model-generating scripts in another file but has the un-commented snippet. This way, I can move the snippet out of the tests. I'll just have to remember that, if we make edits, I'll have to update both the tests and the sample-generating code. And, as you say, we can have Travis run the samples to make sure the sample-generating process is not broken. So should we start a `samples` folder next to the `test` folder?

honeybee_energy/hvac/_base.py
chriswmackey: As I said in response to your first comment, I can take this out this `_HVACEnumeration` class in `_base` and hard code in the fact that `IdealAirSystem` is the only HVAC system that comes with `honeybee_energy`. We will still have a need for this dictionary of `HVAC_TYPES` since this dictionary is where all of the `Room.from_dict` methods look up the HVAC class in order to re-serialize various different HVAC dictionaries. But we'll just have the `honeybee_energy_hvac` extension add classes directly to this dictionary. So this extra step of searching for modules in the `hvac` subpackage can be taken out. Please let me know if you agree.

tests/model_extend_test.py
mostaphaRoudsari: These are two different tasks:

1. Testing the samples for honeybee.
2. Generating sample schemas for honeybee-schema module.

This file is supposed to test extending the model. I think they should be fully separated. Separation of concerns is more important than code duplication.

honeybee_energy/hvac/_base.py
mostaphaRoudsari: I think this is a better approach.

honeybee_energy/writer.py
mostaphaRoudsari: I don't think we should create an exception for ideal air system in this case even if it is 3 lines versus one.

Also I'm not sure why it will be "used so much all over the place". It will be used once as the default system but other than that it is no different from other HVAC systems. I suggest to keep it consistent.

honeybee_energy/hvac/_base.py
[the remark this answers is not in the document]
mostaphaRoudsari: Not sure how this reply is related to the original comment about importing for os.

I believe this is in relation to creating the dictionary which I replied to below.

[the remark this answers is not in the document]
chriswmackey: Hm.  Now I'm less sure what you are commenting on. Were you referring to the `from os.path import dirname, basename, isfile, join` or the `from importlib import import_module`?

I mean, all of this importing is going to get removed because of the dictionary comment but it would still be good to know what you mean here for future reference and for the sake of ladybug.

honeybee_energy/hvac/_base.py
chriswmackey: Done!

tests/model_extend_test.py
chriswmackey: Great! Do you agree with making a `samples` folder next to the `tests` folder to house all of this code that generates the JSON samples for `honeybee-schema`?  Or does code like this actually belong in the `honeybee-schema` repo?

honeybee_energy/writer.py
chriswmackey: It was so nice, though.  Like [here](https://github.com/ladybug-tools/honeybee-grasshopper-core/pull/18/files). It also helped standardize the naming of the default ideal air system (now, we'll have to set the name differently in each of the 3 components there). I mean, nothing will break by not having a standard naming convention, but it was just nice.

chriswmackey:
@mostaphaRoudsari , 
For the sake of not holding up the PRs on other repos that are updating to this change, I am going to merge this one in. I can send some more PRs afterwards to address any other comments you have.

tests/model_extend_test.py
mostaphaRoudsari: > does code like this actually belong in the honeybee-schema repo?

That's a good question. That might actually be the right place to generate the model samples and save them. It also makes sense to have the samples updated in each release.

If you decide to keep them in this repository for now I think the folder is better to be called `scripts`.

honeybee_energy/hvac/_base.py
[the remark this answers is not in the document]
mostaphaRoudsari: I meant mainly for the os module. Not to use `from * import x, y, z`.

tests/model_extend_test.py
chriswmackey: Ok. Let's do it under the honeybee_schema repo, then. [I added an issue](https://github.com/ladybug-tools-in2/honeybee-schema/issues/13) and I'll aim to address it soon after I've finished the other tasks on my sprint.

PR #75, opened 2020-01-27
feat(door): Allow Shades to be assigned to Doors

chriswmackey:
This commit brings the library in line with [this PR on honeybee-core](https://github.com/ladybug-tools/honeybee-core/pull/52).


chriswmackey:
Note that the tests will not pass until the honeybee-core PR is merged in.

PR #82, opened 2020-02-06
feat(cli): Add steps to re-serialize and check the Model

chriswmackey:
This commit adds an option into the current CLI calls that translate honeybee model JSONs to OSM. The option lets CLI users re-serialize and check the model before it goes through the energy_model_measure.

Specifically, the checks ensure that all Faces of a Room volume point outwards, ensure that all Apertures/Doors point in the same direction as their parent Face, and the checks will scale the model to deal with input from different units systems.

@mostaphaRoudsari ,
Given that I am still getting the hang of the CLI stuff, any feedback that you have on this would be helpful.


chriswmackey:
Since I just discovered a pretty big bug that I've now fixed with this PR, I am just going to merge it in. The review here wasn't critical, anyway.

PR #90, opened 2020-02-17
Add dictionary format examples in from_dict()  docstrigs

santiagogaray:
Hi @chriswmackey 
I hope I got these right, I did my best to figure out the keys and differences between from_dict and from_dict_abridged but I feel pretty good.

One minor comment I think the 'layers' key was missing from the WindowConstruction.from_dict() example. I added it, hope that's okay.

Please let me know if I got these right, Note also in same cases I specified 'str' in 'name' fields , you might have better options for some of these.

### DIFF
--- a/honeybee_energy/config.py
+++ b/honeybee_energy/config.py
@@ -121,7 +121,7 @@ class Folders(object):
 
         This folder must have the following sub-folders in order to be valid:
 
-        * ladybug - ruby library with modules for model translation to OpenStudio.
+        * from_honeybee - ruby library with modules for model translation to OpenStudio.
         * measures - folder with the actual measures that run the translation.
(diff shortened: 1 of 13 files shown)


santiagogaray:
Thanks @chriswmackey for the comments. I did the changes and should be ready to merge.

I will continue with honeybee-radiance repo as you suggested. I look forward to get into the wiki and once I finish this one. Also we have a couple of smaller documentation tasks in the contract so we can get to those whenever you are ready.

Also, just wanted to remind you about this PR I did a few days back just in case: https://github.com/ladybug-tools/ladybug-tools-template/pull/13

chriswmackey:
Thank @santiagogaray . Merged!

PR #99, opened 2020-02-22
feat(sql): Add ability to request multiple outputs at once

chriswmackey:


honeybee_energy/result/sql.py
```diff
@@ -93,7 +94,12 @@ def data_collections_by_output_name(self, output_name):
         try:
             # extract all indices in the ReportDataDictionary with the output_name
             c = conn.cursor()
-            c.execute('SELECT * FROM ReportDataDictionary WHERE Name=?', (output_name,))
+            if isinstance(output_name, str):
+                c.execute('SELECT * FROM ReportDataDictionary WHERE Name=?',
+                          (output_name,))
+            else:
+                c.execute('SELECT * FROM ReportDataDictionary WHERE Name IN {}'.format(
+                    output_name))
```
mostaphaRoudsari: I know this PR is already merged but for a better solution that doesn't pose SQL injection risk see here: https://stackoverflow.com/a/1310001/4394669

mostaphaRoudsari: One more comment. I would suggest using `conn.close()` one in `finally`.

Issue #91, opened 2020-02-17
Add a module to load measures, store their arguments, and pass these along to an OSW


chriswmackey:
This will be needed to integrate measures in all of our workflows.

PR #175, opened 2020-05-18
feat(measure): Add a module to load measures and set argument

chriswmackey:
This commit adds a measure module with a class that can parse any OpenStudio measure to a Python object on which the measure arguments can be assigned and then passed to the simulation.

The class in the handles validation of the measure argument inputs and can check to be sure that all required arguments are set before simulation.

There's now an additional input on the run.to_openstuido_osw method such that these measure objects can be included in the .osw files that are passed to the OpenStudio CLI.

I still have to add some honeybee-energy CLI commands that make use of this code but this commit lays the foundation for what Konrad will need to include the reporting measures in his simulations. It is also what will be used to expose measures in the Grasshopper interface.

Regarding this last part, there are two things I wanted you to be aware of, @mostaphaRoudsari :
1. I tested that the native Python xml module works in the GHPython of Rhino 6. However, you might remember that it does not exist in Rhino 5. So, given this current implementation here, measures will only be supported in Grasshopper with Rhino 6. I wrote the code in a way that everything else in Rhino 5 won't be affected and it's just the ability to load measures that will be lacking.
2. It seems like exposing measures in Grasshopper is going to break our rule about not having the number of inputs of a component change. I can't think of any way that I can expose measure arguments that doesn't do this and I just wanted to check with you to see if you are ok with this one exception or to see if you had any other ideas.

Resolves https://github.com/ladybug-tools/honeybee-energy/issues/91


honeybee_energy/measure.py
```diff
@@ -0,0 +1,428 @@
...
+
+    Args:
+        folder: Path to the folder in which the measure exists. This folder
+            must contain a measure.rb and a measure.xml file. Other files are
+            optional.
+
+    Properties:
+        * folder
+        * xml_path
+        * rb_path
```
mostaphaRoudsari, line 20: `measure_path` might be a better option. `rb_path` gives a feeling that this is a ruby installation path to me.

I would also use `args_path` instead of `xml_path`. Basically what the files do rather than what is the file type.

honeybee_energy/measure.py
```diff
@@ -0,0 +1,428 @@
...
+        * identifier
+        * display_name
+        * description
+        * type
+        * arguments
+    """
+
```
mostaphaRoudsari, line 27: It is not required by `__slots__` are missing from this class.

honeybee_energy/measure.py
```diff
@@ -0,0 +1,428 @@
...
+        for atr in root.find('attributes'):
+            if atr.find('name').text == 'Measure Type':
+                self._type = atr.find('value').text
+
+        # parse the measure arguments
+        self._arguments = []
+        for arg in root.find('arguments'):
+            arg_obj = MeasureArgument(arg)
+            if arg_obj.model_dependent:
+                # TODO: Figure out how to implement model-dependent arguments
```
mostaphaRoudsari, line 246: With our current implementation it should be so hard to map the objects from an OpenStudio model to a Honeybee model but I'm looking forward to see what will be your decision on this.

honeybee_energy/measure.py
```diff
@@ -0,0 +1,428 @@
...
+        'String': str,
+        'Choice': str
+    }
+
+    def __init__(self, xml_element):
+        """Initialize MeasureArgument."""
+        # set up the argument value and default value
+        self._value = None  # will be set by user
+        self._default_value = xml_element.find('default_value').text \
+            if xml_element.find('default_value') is not None else None
```
mostaphaRoudsari, line 305:
```suggestion
            if xml_element.find('default_value') is not None
```

honeybee_energy/measure.py
```diff
@@ -0,0 +1,428 @@
...
+        # parse the required properties of the argument
+        self._identifier = xml_element.find('name').text
+        self._type_text = xml_element.find('type').text
+        self._type = self.PYTHON_TYPES[self._type_text]
+        required = xml_element.find('required').text
+        self._required = True if required == 'true' else False
+
+        # parse the optional properties of the argument
+        self._display_name = xml_element.find('display_name').text \
+            if xml_element.find('display_name') is not None else None
```
mostaphaRoudsari, line 316: I believe you don't need `else None` for all these case. If it is not set then it will be None.

honeybee_energy/measure.py
```diff
@@ -0,0 +1,428 @@
...
+    @property
+    def valid_choices(self):
+        """Get a list of text for valid inputs for choice arguments.
+
+        This will be None if the argument type is not Choice.
+        """
+        return self._valid_choices
+
+    def validate(self, raise_exception=True):
+        """Check if the argument will cause errors when run through OpenStudio CLI.
```
mostaphaRoudsari, line 410: I would make this explanation more specific. In this case it looks like the check is to ensure all required arguments have a value. This is clearer than causing errors when run through OpenStudio CLI.

honeybee_energy/measure.py
chriswmackey: `measure_path` is what I would expect the root of the whole measure to be called. I'll pick something more descriptive, though.

chriswmackey: I went with `metadata_file` and `program_file` since this seems to be what they are called in the Measure Writers Guide.

honeybee_energy/measure.py
chriswmackey: I was debating whether I should implement it since we are probably going to be adding more to the class in the future. I'll add it for now.

honeybee_energy/measure.py
chriswmackey: I don't follow. In what workflow are you thinking that a measure like that will be used?

honeybee_energy/measure.py
chriswmackey: I checked and it is needed. You get a `SyntaxError` otherwise.

honeybee_energy/measure.py
chriswmackey: My understanding is that measures that are model dependent are intended for the case where you already have an OpenStudio model, like when you are in OpenStudio GUI. Not like measures that you would include in a workflow like what we primary do with them.

PR #211, opened 2020-06-01
feat(shades): Add support for Blinds and Shades a way that E+ supports

chriswmackey:
This commit might be called a fix as much as it is a new feature. Long story short is that the way that blinds and shades were represented in the schema previously, they could never be simulated by EnergyPlus since E+ always needs a ShadingControl object in order to be able to simulate them.  So I added a new WindowConstructionShaded class to handle the creation of this ShadingControl object and allow users to change some of its properties.

I still have to add tests here and send parallel PRs to honeybee-schema and the honeybee-openstudio-gem but I have verified that it's working in the direct-to-idf case.

@mostaphaRoudsari , I mostly requested your review here to get your thoughts on the name of the new class: `WindowConstructionShaded`.  If my goal is to communicate that this is a version of the `WindowConstruction` class that includes an extra shade layer, I feel that this name is clearer than `ShadedWindowConstruction`. But let me know if you feel otherwise.

Also FYI, while the dynamic behavior of these objects can be mapped to honeybee-radiance, the reverse isn't necessarily true right now and these dynamic construction objects are a lot simpler than the very flexible capabilities we have in honeybee-radiance. At a later point, I will add options in honeybee-energy for more sophisticated dynamic constructions in honeybee-energy that will be map-able back and forth from radiance to energy. At some point, it might also be helpful to add simple dynamic modifiers to honeybee-radiance that mimic what is happening here in honeybee-energy.

### DIFF
--- a/honeybee_energy/construction/_base.py
+++ b/honeybee_energy/construction/_base.py
@@ -32,6 +32,7 @@ class _ConstructionBase(object):
         * u_factor
         * r_factor
         * is_symmetric
+        * has_shade
     """
     # generic air material used to compute indoor film coefficients.
     _air = EnergyWindowMaterialGas('generic air', gas_type='Air')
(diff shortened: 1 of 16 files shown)


chriswmackey:
Thanks for the feedback. I'll change it to `has_shade`. I agree that `Shaded` is misleading but, for the sake of not making the class name long or making it unclear what the `W` stands for, I might just go with `WindowConstructionShade`.

PR #282, opened 2020-07-14
feat(ventcool): Add objects for simple window-based ventilative cooling

chriswmackey:
This commit adds objects for simple window-based ventilative cooling. While this commit only includes enough objects to model windows with ZoneVentilation:WindandStackOpenArea objects, many of these objects should also be used in the future AFN.

@saeranv , Please feel free to review and post any questions that you have as comments. It may take me a day or two before I have enough tests in place to merge this in.  FYI, I am not expecting you to add any `to_idf` method as part of your work (you can see one of them in this PR). I only included this so that we have some way of exporting simple window-based ventilation objects in direct-to-idf workflows. I'm imagining that all of the AFN export happens through the honeybee_openstudio_gem.

### DIFF
--- a/honeybee_energy/properties/aperture.py
+++ b/honeybee_energy/properties/aperture.py
@@ -3,6 +3,7 @@
 from ..construction.dictutil import dict_to_construction
 from ..construction.window import WindowConstruction
 from ..construction.windowshade import WindowConstructionShade
+from ..ventcool.opening import VentilationOpening
 from ..lib.constructionsets import generic_construction_set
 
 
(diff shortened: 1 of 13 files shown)


saeranv:
Thanks @chriswmackey, this helps understand how the AFN module will look like. 

I think I'll have more questions as I get deeper into coding, but right now everything you wrote makes sense. 

One clarification: Are the VentilationControl and VentilationOpening the same objects I'm using for the AFN, i.e they reference the AFNRoom and the AFNOpening in my this diagram? 

image


So the idea is the  ZoneVentilation:WindandStackOpenArea will now have the same objects for control/opening as the AFN objects. 

image


So I should be modifying the VentilationOpening and VentilationControl objects when I add the additional parameters for the AFN?

chriswmackey:
@saeranv ,

My instinct tells me that the `VentilationControl` and the `VentilationOpening` objects should be used in both the AFN and the simple ventilation workflows. It just doesn't feel right to have two different places where the user specifies the fraction of the Aperture that is operable or the Room setpoint at which the windows open. So I think you should just extend these two objects with whatever additional properties you thing are needed for the AFN.

Also FYI, I have been thinking about the SimulationContol and I feel that both the `ReferenceCrack` and the "toggle" to switch between the use of AFN vs. simple airflow objects should live on the `ModelEnergyProperties` and not the `SimulationParameter` object. This really is more of a property of the building design rather than a simulation setting and putting these objects in `SimulationParameter` would make it near-impossible to run the honeybee-openstudio-gem without simulation parameters (something that can be done right now without issues).

PR #304, opened 2020-07-27
feat(afn): Add AFN objs and modify corresponding ModelEnergyProperties

saeranv:
@chriswmackey , @mostaphaRoudsari 

This PR contains the honeybee-energy AFN objects. For the most part it is a straightforward translation of the equivalent honeybee-schema objects. Some things to pay attention to:

- I debated on the best way to implement the VentilationSimulationControl (VSC) object in the ModelEnergyProperties. All the other ModelEnergyProperties are automatically extracted from the host Model passed in the constructor. Since the VSC object isn't something we can extract from the host Model, it is therefore treated differently from the other ModelEnergyProperties. Specifically, it's the only property that is settable, and the only property you can pass as an argument in the constructor. If nothing is passed, then a default VSC object will be made for the ModelEnergyProperties. My reasoning for this is that the default of SingleZone ventilation cooling should be able to model wind and stack airflow for any Honeybee model, with very little additional work, so it's more convenient to have it set there by default. Alternatively I can remove it from the constructor and just keep it as an optional get/set property. Thoughts?
- I am currently working on a method to auto-generate the AFN from a list of rooms, but didn't commit it in this PR, for the sake of clarity. 
- The way the VentilationSimulationControl object handles reference building type data may be modified, based on decisions we're making in my Honeybee-schema PR regarding the handling of complex geometries.

This is draft PR for now since it's missing some methods, but I think it's ready for a review. 

Thanks


chriswmackey:
Thanks @saeranv . I will review this tomorrow.

honeybee_energy/ventcool/crack.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Properties:
+        * air_mass_flow_coefficient_reference
+        * air_mass_flow_exponent
+        * crack_factor
+        * parent
+        * has_parent
+    """
+    __slots__ = ('_air_mass_flow_coefficient_reference', '_air_mass_flow_exponent',
+                 '_crack_factor', '_parent', '_locked')
```
chriswmackey, line 41: As I said, I don't think you need the whole parent capability here.

honeybee_energy/ventcool/crack.py
```diff
@@ -0,0 +1,165 @@
...
+            0.5 generally corresponding to turbulent orifice flow and 1 generally
+            corresponding to laminar flow. The default of 0.65 is representative of many
+            cases of wall and window leakage, used when the exponent cannot be measured.
+            (Default: 0.65).
+        crack_factor: A number indicating multiplier for air mass flow through a crack.
+            (Default: 1).
+
+    Properties:
+        * air_mass_flow_coefficient_reference
+        * air_mass_flow_exponent
```
chriswmackey, line 35: These are really long variable names. Can we just call them `flow_coefficient` and `flow_exponent`?  Generally speaking, it gets hard to write code with variables that are longer than 25 characters since these take up 1/3 of the whole character limit in a line.

honeybee_energy/ventcool/crack.py
```diff
@@ -0,0 +1,165 @@
...
+    def __init__(self, air_mass_flow_coefficient_reference, air_mass_flow_exponent=0.65,
+                 crack_factor=1):
+        """Initialize AFNCrack."""
+        self.air_mass_flow_coefficient_reference = air_mass_flow_coefficient_reference
+        self.air_mass_flow_exponent = air_mass_flow_exponent
+        self.crack_factor = crack_factor
+        self._parent = None  # this will be set when assigned to an aperture
+
+    @property
+    def air_mass_flow_coefficient_reference(self):
```
chriswmackey, line 52: Yea, this is way too long of a variable name. Let's use the docstring if we want to add more adjectives to the description like "reference"

honeybee_energy/ventcool/crack.py
```diff
@@ -0,0 +1,165 @@
...
+        """Get or set the air mass flow coefficient defined at the reference crack."""
+        return self._air_mass_flow_coefficient_reference
+
+    @air_mass_flow_coefficient_reference.setter
+    def air_mass_flow_coefficient_reference(self, value):
+        assert value > 0, 'The air_mass_flow_coefficient_reference must be greater ' \
+            'than 0. Got: {}.'.format(value)
```
chriswmackey, line 59: This isn't the right way to do range checks in honeybee since there are a bunch of other cases where people input strings, etc. See [how it's done in the VentilationOpening class here](https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/ventcool/opening.py#L66).

chriswmackey: Or, in this case, you probably want to use the `float_positive` check that is in the same honeybee-core module.

honeybee_energy/ventcool/crack.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @property
+    def air_mass_flow_exponent(self):
+        """Get or set the air mass flow exponent for the surface crack."""
+        return self._air_mass_flow_exponent
+
+    @air_mass_flow_exponent.setter
+    def air_mass_flow_exponent(self, value):
+        assert 1 >= value >= 0.5, 'The air_mass_flow_exponent must be greater ' \
+            'than or equal to 0.5 and less than or equal to 1. Got: {}.'.format(value)
```
chriswmackey, line 70: Same comment here.

honeybee_energy/ventcool/crack.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @property
+    def crack_factor(self):
+        """Get or set the multiplier for air mass flow through a crack."""
+        return self._crack_factor
+
+    @crack_factor.setter
+    def crack_factor(self, value):
+        assert 0 < value <= 1, 'The crack_factor must be greater ' \
+            'than 0 and less than or equal to 1. Got: {}.'.format(value)
```
chriswmackey, line 81: I'll just let you fix all of these throughout your code without further comments here.

honeybee_energy/ventcool/crack.py
```diff
@@ -0,0 +1,165 @@
...
+        crack_fac = data['crack_factor'] if 'crack_factor' in data \
+            and data['crack_factor'] is not None else 1
+
+        return cls(air_coeff, air_exp, crack_fac)
+
+    def to_dict(self):
+        """AFNCrack dictionary representation."""
+        base = {'type': 'AFNCrack'}
+        base['air_mass_flow_coefficient_reference'] = \
+            self.air_mass_flow_coefficient_reference
```
chriswmackey, line 129: Case in point about big variable names.

honeybee_energy/ventcool/opening.py
```diff
@@ -34,25 +34,59 @@ class VentilationOpening(object):
             cross ventilation will be induced. If False, the assumption is that
             the operable area is primarily on one side of the Room and there is
             no wind-driven ventilation. (Default: False)
+        air_mass_flow_coefficient_closed: A number in kg/s-m, at 1 Pa per meter of crack
```
chriswmackey, line 37: what about `flow_coefficient_closed`?

honeybee_energy/ventcool/opening.py
```diff
@@ -34,25 +34,59 @@ class VentilationOpening(object):
...
+            required to run an AirflowNetwork simulation. Some common values for this
+            coefficient from the DesignBuilder Cracks template include the following:
+
+                * 0.00001 - Tight, low-leakage closed external window
+                * 0.003 - Very poor, high-leakage closed external window
+
+            If this value is None and one of the AirflowNetwork options are
+            selected in the VentilationSimulationControl object, a default value will be
+            calculated to produce air flow leakages equivalent to the zone flow rate
+            defined by the corresponding Infiltration object.
```
chriswmackey, line 48: No, I don't think that's not what is going to happen here with a `None` input. Where do you propose this auto-calculation code is going to be? And how do you even know that a Face has a parent Room with an infiltration object? Just set it to 0 by default and you'll add a method on the RoomEnergyProperties that users can explicitly call to compute the flow coefficient this from the Room's infiltration object if it exists.

Or, maybe a wiser choice is always to leave this at 0 for such cases since you can probably account for the whole Room infiltration with just the parent Face'c crack.

Also, you should state the default value here in the docstring.

honeybee_energy/ventcool/opening.py
```diff
@@ -34,25 +34,59 @@ class VentilationOpening(object):
...
+        air_mass_flow_exponent_closed: An optional dimensionless number between 0.5 and
+            1 used to calculate the mass flow rate when the opening is closed; required
+            to run an AirflowNetwork simulation. This value represents the leak geometry
+            impact on airflow, with 0.5 generally corresponding to turbulent orifice flow
+            and 1 generally corresponding to laminar flow. The default of 0.7 is
+            representative of internal and external window leakage, according to the
+            DesignBuilder Cracks template.
+        minimum_density_difference_two_way: A number in kg/m3 indicating the minimum
```
chriswmackey, line 56: Way too long of a variable name. What about `two_way_threshold`.

honeybee_energy/ventcool/opening.py
```diff
@@ -209,6 +289,10 @@ def to_dict(self):
         base['fraction_height_operable'] = self.fraction_height_operable
         base['discharge_coefficient'] = self.discharge_coefficient
         base['wind_cross_vent'] = self.wind_cross_vent
+        base['air_mass_flow_coefficient_closed'] = self.air_mass_flow_coefficient_closed
```
chriswmackey, line 292: Maybe you want to do a check to see if this is not 0 before writing all of the AFN stuff into the dict.

honeybee_energy/ventcool/opening.py
```diff
@@ -34,25 +34,59 @@ class VentilationOpening(object):
             cross ventilation will be induced. If False, the assumption is that
             the operable area is primarily on one side of the Room and there is
             no wind-driven ventilation. (Default: False)
+        air_mass_flow_coefficient_closed: A number in kg/s-m, at 1 Pa per meter of crack
+            length, used to calculate the mass flow rate when the opening is closed;
+            required to run an AirflowNetwork simulation. Some common values for this
```
chriswmackey, line 39: Change this from "required to run an AirflowNetwork simulation" to "only used in AirflowNetwork simulation."  The default should be 0 if it's not specified.

honeybee_energy/ventcool/opening.py
```diff
@@ -129,21 +196,34 @@ def from_dict(cls, data):
...
         """
         assert data['type'] == 'VentilationOpening', \
             'Expected VentilationOpening dictionary. Got {}.'.format(data['type'])
 
         area_op = data['fraction_area_operable'] if 'fraction_area_operable' in data \
             and data['fraction_area_operable'] is not None else 0.5
-        height_op = data['fraction_height_operable'] if 'fraction_height_operable' in data \
-            and data['fraction_height_operable'] is not None else 1.0
+        height_op = data['fraction_height_operable'] if 'fraction_height_operable' in \
+            data and data['fraction_height_operable'] is not None else 1.0
```
chriswmackey, line 210: Did you check to be sure the default discharge coefficient of 0.17 makes sense for the AFN?  I know that I originally interpreted this assuming the `fraction_height_operable` is the full height of the window (and the discharge coefficient is reduced as a result) but I don't know if the AFN makes the same assumption.

honeybee_energy/ventcool/simulation_control.py
```diff
@@ -0,0 +1,261 @@
...
+
+            The MultiZone options will model air flow with the AirflowNetwork model,
+            which is generally more accurate then the SingleZone option, but requires
+            defining more ventilation parameters to explicitly account for weather
+            and building-induced pressure differences, and the leakage geometry
+            corresponding to specific windows, doors, and surface cracks.
+            (Default: 'SingleZone').
+        reference_temperature: Reference temperature measurement in Celsius under which
+            the surface crack data were obtained. (Default: 20).
+        reference_barometric_pressure: Reference barometric pressure measurement in
```
chriswmackey, line 29: `reference_pressure` is descriptive enough for me.

honeybee_energy/ventcool/simulation_control.py
```diff
@@ -0,0 +1,261 @@
...
+
+    @property
+    def reference_temperature(self):
+        """Get or set the temperature for the reference crack."""
+        return self._reference_temperature
+
+    @reference_temperature.setter
+    def reference_temperature(self, value):
+        assert isinstance(value, (int, float)), 'reference_temperature must be ' \
+            'a number. Got: {}.'.format(value)
```
chriswmackey, line 113: You know how to fix all of these now so you don't have to do a much of checks. You probably want to make sure this one is above absolute zero.

saeranv:
Update: Now this PR has the functions to auto-generate an AFN from a list of rooms. I need to modify this function to be a little bit more user-friendly, but wanted to get it in here early so it can be included in any review.

saeranv:
@chriswmackey  with this update the afn.generate function is also ready for review.

honeybee_energy/properties/face.py
```diff
@@ -67,6 +72,28 @@ def construction(self, value):
...
+                'for Face vent_crack. Got {}'.format(type(value))
+            if value._parent is None:
+                value._parent = self.host
+            elif value._parent.identifier != self.host.identifier:
+                raise ValueError(
+                    '{0} objects can be assigned to only one parent.\n{0} cannot be '
+                    'assigned to Face "{1}" since it is already assigned to "{2}".\n'
+                    'Try duplicating the object and then assign it.'.format(
+                        'AFNCrack', self.host.identifier,
+                        value._parent.identifier))
```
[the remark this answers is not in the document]
saeranv: @chriswmackey 

I thought it would be useful to treat the cracks the same as the vent openings, but on second thought I think this makes more sense. The AFNCrack won't change for different host geometries or identifiers, but it does change based on surface type, boundary condition, and between rooms, but I think locking and using the same crack will be fine for each of these groups.

honeybee_energy/properties/model.py
```diff
@@ -442,17 +459,25 @@ def to_dict(self):
...
         self._add_sched_type_objs_to_dict(base, schs)
 
+        # add ventilation_simulation_control
+        base['energy']['ventilation_simulation_control'] = \
+            self.ventilation_simulation_control.to_dict()
+
         return base
 
-    def duplicate(self, new_host=None):
+    def duplicate(self, new_host=None, new_ventilation_simulation_control=None):
```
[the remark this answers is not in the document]
saeranv: Makes sense, thanks.

honeybee_energy/ventcool/crack.py
saeranv: Yeah I agree, I just suck at variable naming. Will change.

honeybee_energy/ventcool/crack.py
saeranv: @chriswmackey 

This one actually may pose a problem. In this case the lower bound is greater than 0 but not equal to 0, which is not possible to achieve with the float_in_range function (since it always assumes inclusive ranges (greater/less then or equal to). 

I can fudge this by using an extremely small value for the lower range (0.0000001) or add three new functions in honeybee.typing to account for these kinds of range checks. What do you think?

saeranv: I think the best thing to do is add three new functions to account for different range checks. I'll include it in this module and move it to honeybee.typing if it looks good.

chriswmackey:
Thanks, @saeranv ,
Can you re-request a review from me whenever you want me to look over this all again?  It's a little unclear if all of the previous comments have been addressed yet.

honeybee_energy/ventcool/opening.py
saeranv: @chriswmackey 

**Re: setting flow_coefficient to 0 as default**
Technically the flow_coefficient_closed value has to be greater than 0 (there's a float_positive check to ensure this). If the goal is to get rid of None as a default, I think there are three alternatives:

1. Allow a zero value. This isn't ideal because it could cause confusion if we allow 0 here (even if we are planning on overwriting it), and EnergyPlus doesn't.  The flow_exponent_exponent can't be set to zero either since that'll make the flow rate equal to 1 and there's a minimum value of 0.5. 

2. Set the flow_coefficient to some assumed default based on Face type and boundary condition. The flow_coefficient for openings has a range from 0.00001 to 0.04 (from DesignBuilder) based on if it's an interior/exterior window/door, so we'll have to apply this default based on the host aperture properties. I can include a private method in vent_opening that extracts this value from the host. Then this default (which will have to be based on the DB crack template references) can be overwritten by the auto-calc method in the RoomEnergyProperties if the user wants it to simulate the zone infiltration flow rate.

3. Add a very small value as a default like 0.00001. So by default we represent a very tight envelope. The problem with this is that it doesn't correlate with the default envelope properties for different Face types/boundary conditions. So it's a unneccessarily inaccurate default. 

I'm like option number 2, but am open to another options.

**Re: Accounting the whole Room infiltration with just the parent Face crack.**  
Right now I'm translating the zone infiltration for leakage through opening edges AND surface area cracks:
https://github.com/ladybug-tools/honeybee-energy/pull/304/files#diff-cd4564ea076fa96e5cb3a05f5a09915eR248-R278. Since the blower door tests used to derive the zone infiltration rates were done one buildings with a mix of surface cracks and opening edges, this will give us relatively more accurate surface/edge opening parameters. In terms of the resulting infiltration load there shouldn't be any difference between these approaches, but it could be misleading if someone were looking at the Face crack coefficients.  

**Re: RoomEnergyProperties auto-calculation from infiltration**
As for the auto-calculation, right now it's just a function in the afn.py module (you can see it above if you scroll up), but I don't see any problem moving the code to RoomEnergyProperties.

honeybee_energy/ventcool/opening.py
saeranv: Will revisit this once we figure out default assumptions from discussion above.

honeybee_energy/ventcool/opening.py
saeranv: @chriswmackey 

Can you clarify how you calculate the default discharge coefficient?

There's no default in EP. Here's what I found from other sources:
- DesignBuilder suggests 0.65 as a "good general purpose setting". 
- CONTAMW (which the EP engineering reference cites) suggests "0.6 for an orifice and slightly higher for other openings in buildings" (so 0.65 seems reasonable for vent_opening). 
- EP sample files has 0.5 for closed opening and 0.6 for completely open window. 

So I could change it to 0.6+ and cite CONTAMW. But let me know how you calculated 0.17 in case there's something else we should account for. 

I should add, since the discharge coefficient is the ratio between actual/ideal air flow, and EP will calculate the ideal flow in the ZoneVentilation and AFN in comparable ways (former with wind/stack kinetic/gravitational potential energy, and latter with pressure) I don't see why they should be radically different.

chriswmackey:
I see that you have some questions for me here but you didn't re-request my review. I'll assume this is ready for another pass.

honeybee_energy/ventcool/crack.py
chriswmackey: I think that's fine if you want to add new functions to honeybee-core, though I think we were finding in your other honeybee-schema PR that we probably want to remove the `crack_factor` and just have people edit the `flow_coefficient`.  So this particular example doesn't require it.

If there are other range checks where you need to exclude the upper/lower value, then you can still add the methods to honeybee-core. I think we should only be strict with this in cases where EnergyPlus fails if a variable is at a certain value, though.

saeranv:
> I see that you have some questions for me here but you didn't re-request my review. I'll assume this is ready for another pass.

Yes, I was hoping to migrate some afn.py methods to RoomEnergyProperties before requesting a full review, but answering those questions will be helpful.

honeybee_energy/ventcool/opening.py
chriswmackey: I am definitely leaning towards option 1 as the other two seem to be making some dangerous implicit assumptions for the user without their awareness.  And having a default of no infiltration is consistent with how everything else in honeybee works. For example, if someone initializes an empty `ProgramType` object, the object has no infiltration until the user specifies it. The same thing should be true of the AFN and the user should have to explicitly specify the flow coefficient or run some method to autocalculate it based on other properties.

Whatever the case, a flow coefficient of 0 should definitely be permissible since there has to be some way for the user to switch off the connection of a certain object to the AFN if they want to.  And bear in mind that this default of 0 isn't something that gets exposed on the Grasshopper components (we'll always run the autocalculation for the user if they request to use the AFN). It's just that, on the Python SDK layer we should be as explicit as possible.

Intuitively, it feels like the Room auto-calculation should be method on `RoomEnergyProperties` since the method doesn't get applied to anything else.  I'll take a look over it now.

honeybee_energy/ventcool/opening.py
chriswmackey: As I said above, there needs to be some way for people to disconnect the object from the AFN if they don't want the opening to have crack flow through it. So supporting 0 is very useful.

I'll also give another reason why we can't use `None` in the schema:
The C# bindings for the schema have much stricter type enforcement than the Python bindings. So this means that having both `None` and a float value is not acceptable. If there really was some way to have these values autocalculated, we should use a union of float with the [honeybee-core Autocalculate AltNumber](https://github.com/ladybug-tools/honeybee-core/blob/master/honeybee/altnumber.py#L37-L43). But I have been reserving this `Autocalculate` object  just for cases where EnergyPlus has an autocalculate feature. So, again, I think 0 is the best default here.

honeybee_energy/ventcool/opening.py
chriswmackey: So what I am hearing from you is that the discharge coefficient is NOT supposed to be computed using the full height of the window whenever the AFN is used.  Instead, it assumes that the "height" is the average distance from the center of the opening to the neutral plane (or roughly a quarter of the window's height). I know this because 0.65 is the generally agreed-upon discharge coefficient that should be used when windows are unobstructed (no insect screens) and the height is a quarter of the window height.

I guess this 0.65 is what most people use but it always just seemed really un-intuitive for me to use a quarter of the window height and it was much better to wrap this assumption up with the coefficient.  But so be it. I'll plan to change the honeybee-opensutdio-gem to quarter the height of the window for simple ventilation (we can still keep the `fraction_height_operable` input that we use in the schema) and I'll change every other place where the discharge coefficient is used to be representative of this quartering of the height.

Also, I know that the discharge coefficient for a quarter-height window opening AND insect screens is 0.45.  So what do you think about making the default discharge coefficient here to be 0.45 instead of 0.65?  It seemed better to air on the side of conservative air flow but I know I'm also biased by living in a climate where insect screens are a must-have.

chriswmackey:
Thank you for letting me know @saeranv . I can hold off on a fell review of the new methods until you put them on the `RoomEnergyProperties`, then.  Generally, I like how well you documented everything but I think some of the `solve_` functions can just be `staticmethods` on the properties class.  I also think that you don't need to pull out the constants like the density of air into their own variables and you can just set them as defaults on these methods.  The namedtuples can just be constant variables on the RoomEnergyProperties class (with UPPERCASE styling).

Lastly, I think we need to avoid passing the adjacency info into the method that builds the AFN since this is not going to be readily available in a lot of cases. People may also be deconstructing the model into separate Rooms and adjacent surfaces should each keep a record of the crack flow across their interior surfaces. I realize that this can expose us to cases where the AFN properties don't match between adjacent surfaces and maybe we need to add an extra check for this but such a check is much preferable to a case where the model cannot be deconstructed and an adjacent Room can be replaced with a new one.

We can take care of ensuring that only one crack gets written into the IDF for each pair of adjacent surfaces in the honeybee-openstudio-gem. So I would say don't worry about this case here.

honeybee_energy/ventcool/opening.py
chriswmackey: I think I'll make this change toady and I am going to make the default 0.45 in order to be conservative with the insect screens.

saeranv: @chriswmackey thanks for explaining this. I agree with the 0.45 value since it's much closer to the established precedent of 0.6 - 0.65, (and I agree we should reduce it to account for the prevalence of insect screens).

Furthermore, having done more research into the discharge coefficient, I found this quote: "For example IEA Annexe 20 Air flow Through Large Openings in Buildings indicates that discharge coefficients reported in the literature vary between 0.3 and 0.8 and that it is not understood what causes the differences" (from DesignBuilder docs), which suggests 0.45 is well within acceptable range, and there's a lot of uncertainty associated with this value anyway. 

Regarding this the window height:
> So what I am hearing from you is that the discharge coefficient is NOT supposed to be computed using the full height of the window whenever the AFN is used. Instead, it assumes that the "height" is the average distance from the center of the opening to the neutral plane (or roughly a quarter of the window's height). I know this because 0.65 is the generally agreed-upon discharge coefficient that should be used when windows are unobstructed (no insect screens) and the height is a quarter of the window height.

I've been trying to find more information on the relationship between the discharge and window/neutral pressure plane heights, and so far I can't find any confirmation of this. 

The EP AFN example files assumes a 0.5 'Window/Door Opening Factor' (the % of the window surface that has an opening), and then assumes an 'Opening/Height/Width Factor' (window fully open) of 1 for the discharge coefficient of 0.6, and an 'Opening/Height/Width Factor' of 0 (window fully closed) for the discharge coefficient of 0.5. So it does seem to be assuming 0.6 for a fully open window opening (which is 50% of the window surface). 

I don't think this assumption effects our model at any rate, since the Simple Window Opening object we are using in the AFN only uses one discharge value, and doesn't require different assumptions for different opening heights. And as I said, the 0.45 is reasonable according to multiple reference I've listed above.

honeybee_energy/ventcool/opening.py
saeranv: Okay, and the inability to translate the optional None to C# argument you make below is convincing. I'll switch all these to 0, and document our assumptions.

honeybee_energy/ventcool/opening.py
saeranv: Got it, will switch to 0, and document our assumptions.

honeybee_energy/ventcool/opening.py
saeranv: FYI, I found this reference in the DB docs that helped me understand that the discharge coefficient is actually closely related to the flow coefficient we're using in the Crack model. Specifically, it's based on a different variation of flow rates derived from Bernoulli's equation: 

image

So you can see there that in the orifice equation used for window openings, the 'flow exponent' is square root (0.5), and the Cd*A is serving the same role as the flow-coefficient.

saeranv:
@chriswmackey this is finally ready for review, sorry for the delay! 

Notes:

- I decided to change the Crack template data from namedtuples to a dictionary. I thought this would be consistent with the patterns we're using for honeybee-schema, and also avoids the import of an extra module I need for the NamedTuple module. Let me know if you prefer to change it back. 

- I removed the adjacency data requirement for the interior AFN generation function. You mentioned that we should add a check to confirm that adjacent faces are actually equivalent. Is the place for the check in the gem? Or should I be adding it on my end? 

- I added the method to compute air density from a user-provided atmospheric pressure and the ideal gas law. 

- I included an option in afn.generate to use the crack template to provide all air leakage values, and avoid using the room infiltration derivation method. I also expanded the crack template to include exterior leakage data to be able to do this. I explained the advantages/disadvantages of both in the docstring, but basically, deriving the leakage parameters from the infiltration rate results in generic leakage parameters since it's derived from a single average value. The crack template data in contrast contains a broader range of value representing empirically obtained leakage parameters for different surface types. Thus, the crack template data will allow users to more accurately model the geometric impact of windows/cracks etc. I'll provide a write-up in basecamp about this decision so that we can discuss it more. 

- You'll notice that I am no longer referencing a single AFNCrack for multiple walls. This is because, I was actually making a mistake before. The flow_coefficient values are actually normalized by area, and so technically they need to be multiplied by wall area - thus every wall requires a different AFNCrack. This is how DB does it. Since the AFN is varying the wind pressure across different walls, meaning different walls will produce different flow rates - this can be significantly more accurate then using a single generic AFNCrack. However, I definitely think we could simplify this in the room_infiltration method (just use a single representative wall and window, as you mentioned before) because it's already making assumptions that cancel out any gains in precision.    


S

saeranv:
@chriswmackey 

FYI: the tests here are failing because they are require new float_in_range functions that haven't been merged into honeybee-core yet. The tests are working on my local so I don't forsee any issues once my PR for honeybee-core gets merged.

honeybee_energy/ventcool/opening.py
chriswmackey: The comment I made about the quartering of the window height had more to do with how we should be translating the data about the discharge coefficient for the [simple WindAndStack ventilation objects](https://bigladdersoftware.com/epx/docs/9-3/input-output-reference/group-airflow.html#zoneventilationwindandstackopenarea).  We just want the user input to be consistent between the AFN and the simpler ZoneVentilation where possible.  You can see at the link there that the E+ Input/Output reference basically interprets the discharge coefficient in a consistent manner between AFN and zone ventilation, suggesting that I use the "height from midpoint of lower opening to the neutral pressure level" for the simple objects (the quarter height).  So I made the change in the honeybee-openstudio-gem as we translate the schema input over to OSM format:
https://github.com/ladybug-tools/honeybee-openstudio-gem/pull/134

Because the AFN is much more complex in the the way it models stack effects, we shouldn't have to do this process of quartering the height.  But all of this means that we can keep a consistent definition of discharge coefficients around the 0.3-0.8 range.

honeybee_energy/properties/room.py
```diff
@@ -279,6 +286,200 @@ def is_conditioned(self):
...
+
+        This will compute air leakage parameters for exterior cracks and opening edges
+        that produce a total air flow rate equivalent to the room infiltration rate, at
+        an envelope pressure difference of 4 Pa. However, the individual flow air
+        leakage parameters are not meant to be representative of real values, since the
+        infiltration flow rate is an average of the actual, variable surface flow
+        dynamics.
+
+        VentilationOpening objects will be added to Aperture and Door objects if not
+        already defined. If already defined, only the parameters defining leakage when
```
chriswmackey, line 438: It feels like you are removing a huge degree of freedom here if you are forcing the user to accept that all apertures suddenly become operable when this method is called.  I realize you may still need the VentilationOpening object even when the window doesn't open. But, in this case, can you ensure that the operable area of the VentilationOpening  is 0 so that we aren't totally changing the user's original assumptions here?

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+"""Functions to generate airflownetwork for a list of rooms."""
+from __future__ import division
+import math
+
+from .crack import AFNCrack
+from .opening import VentilationOpening
+from .crack_data import crack_data_dict
+
+
+# TODO: Move to ladybug_geometry.Base2DIn3D?
```
chriswmackey, line 11: Yes, we definitely should or it should be in honeybee-core much like this exact same method currently exists in dragonfly-core:
https://github.com/ladybug-tools/dragonfly-core/blob/master/dragonfly/_base.py#L82-L114

Whatever the case, it should not be a publicly-exposed method because it's not going to live in this spot for long and we don't want other developers using it. So please change them to `_compute_bounding_box_x` and `_compute_bounding_box_y` for now.

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+        if geom.max.z > max_z:
+            max_z = geom.max.z
+        if geom.min.z < min_z:
+            min_z = geom.min.z
+
+    return min_z, max_z
+
+
+# TODO: Move to ladybug_geometry.Base2DIn3D?
+def compute_bounding_box_extents(geometries, axis_angle=0):
```
chriswmackey, line 75: The same goes for all of these. If you want to just leave them here for now, then please hide them from the public SDK.

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+    zdist = zz
+
+    return 'LowRise' if zdist <= hdist else 'HighRise'
+
+
+def compute_aspect_ratio(bounding_box_extents):
+    """Compute the AirflowNetwork aspect ratio of a building.
+
+    Args:
+        bounding_box_extents: # TBD
```
chriswmackey, line 126: What does TBD mean here?  It looks like your'e expecting a real input here.

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+            * int_apertures: List of interior Aperture subface Face objects.
+            * int_doors: List of interior Door subface Face objects.
+
+        int_cracks: A dictionary of air mass flow coefficient and exponent data
+            corresponding to the face types in the interior_face_groups. Face
+            data flow coefficients should be normalized by surface area, and closed
+            opening flow coefficients should be normalized by edge lengths, for example:
+
+            {
+                # kg/s/m2 @ 1 Pa
```
chriswmackey, line 184: Not a correct dosctring.  See all of the other places where we include sample dictionaries in docstrings.

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+    reflects the leakage characteristics of faces with different areas and types.
+    This requires multiplying the area-normalized air mass flow coefficients from the
+    reference crack data in int_cracks, by the wall area.
+
+    This function does not check adjacency information when computing leakage parameters
+    for adjacent faces or subfaces. It assumes adjacent faces share the same type and
+    area properties so that the computed leakage parameters are equivalent. Note that
+    EnergyPlus only requires one of the adjacent interzone surfaces to be
+    assigned a leakage component. If both adjacent surfaces have a leakage component,
+    the air flow through the surface will be counted twice.
```
chriswmackey, line 167: No need for this last sentence or really for this whole paragraph.  As I said in earlier comments, we'll take care of ensuring the IDF gets only one object per adjacent pair as part of the honeybee-openstudio-gem.

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+        face_area = int_floorceiling.area - opening_area
+        flow_cof = int_cracks['floorceiling_flow_cof'] * face_area
+        flow_exp = int_cracks['floorceiling_flow_exp']
+        int_floorceiling.properties.energy.vent_crack = AFNCrack(flow_cof, flow_exp)
+
+    # add interior opening leakage components
+    for int_aperture in int_apertures:
+        if int_aperture.properties.energy.vent_opening is None:
+            int_aperture.is_operable = True
+            int_aperture.properties.energy.vent_opening = VentilationOpening()
```
chriswmackey, line 220: Why are you removing the degree of freedom here?  If you need a VentilationOpening just to set infiltration, then set the operable area to 0.

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+        vent_opening = int_aperture.properties.energy.vent_opening
+        vent_opening.flow_coefficient_closed = int_cracks['window_flow_cof']
+        vent_opening.flow_exponent_closed = int_cracks['window_flow_exp']
+
+    for int_door in int_doors:
+        if int_door.properties.energy.vent_opening is None:
+            int_door.properties.energy.vent_opening = VentilationOpening()
```
chriswmackey, line 227: Same comment here.

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+        face_area = ext_roof.area - opening_area
+        flow_cof = ext_cracks['roof_flow_cof'] * face_area
+        flow_exp = ext_cracks['roof_flow_exp']
+        ext_roof.properties.energy.vent_crack = AFNCrack(flow_cof, flow_exp)
+
+    # add exterior opening leakage components
+    for ext_aperture in ext_apertures:
+        if ext_aperture.properties.energy.vent_opening is None:
+            ext_aperture.is_operable = True
+            ext_aperture.properties.energy.vent_opening = VentilationOpening()
```
chriswmackey, line 291: same here

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+        vent_opening = ext_aperture.properties.energy.vent_opening
+        vent_opening.flow_coefficient_closed = ext_cracks['window_flow_cof']
+        vent_opening.flow_exponent_closed = ext_cracks['window_flow_exp']
+
+    for ext_door in ext_doors:
+        if ext_door.properties.energy.vent_opening is None:
+            ext_door.properties.energy.vent_opening = VentilationOpening()
```
chriswmackey, line 298: same here

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+        vent_opening = ext_door.properties.energy.vent_opening
+        vent_opening.flow_coefficient_closed = ext_cracks['door_flow_cof']
+        vent_opening.flow_exponent_closed = ext_cracks['door_flow_exp']
+
+
+def generate(rooms, window_vent_controls, leakage_type='Average',
+             use_room_infiltration=True, atmospheric_pressure=101325):
```
chriswmackey, line 305: I'm thinking this should be the only public method in this whole `afn.py` module.

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+    VentilationOpening objects will be added to Aperture and Door objects if not already
+    defined. If already defined, only the parameters defining leakage when the openings
+    are closed will be overwritten. AFNCrack objects will be added to all external and
+    internal Face objects, and any existing AFNCrack objects will be overwritten.
+
+    Args:
+        rooms: List of Honeybee Room objects that make up the Airflow Network. The
+            adjacencies of these rooms must be solved.
+        window_vent_controls: List or tuple of VentilationControl objects or None,
+            corresponding to the list of rooms.
```
chriswmackey, line 322: I think we should get rid of this input given that rooms can already have `VentilationControl` objects that are assignable to them.  It's just confusing that you are having people specify VentilationControl objects a second time when someone already has a Model with these objects.

honeybee_energy/ventcool/afn.py
```diff
@@ -0,0 +1,391 @@
...
+        leakage_type: Text identifying the leakiness of the internal walls. This
+            will be used to determine the air mass flow rate parameters for cracks in
+            internal floors, ceilings and walls. (Default: 'Average').
+            Choose from the following:
+
+            * Tight
+            * Average
+            * Leaky
```
chriswmackey, line 330: I assume this comes from the Design Builder templates?  If so, can you add a `Note:` section with a reference like we do in the ladybug-comfort docstrings:
https://github.com/ladybug-tools/ladybug-comfort/blob/master/ladybug_comfort/pmv.py#L19-L30

honeybee_energy/ventcool/crack_data.py
```diff
@@ -0,0 +1,89 @@
+# coding':utf-8
+"""Crack data dictionary.
+
+This data is obtained from the DesignBuilder Crack Template. Note that this
+crack data is normalized by area and perimeter for face cracks and subface edges,
+respectively. In EP face cracks are not normalized by area, but subface edges
+are normalized by perimeter.
+"""
```
chriswmackey, line 8: Can we hide this whole module by calling it `_crack_data.py`?
If there are no methods or classes in the module, then there's no reason for the user to interact with it as part of the public SDK.

honeybee_energy/ventcool/crack_data.py
```diff
@@ -0,0 +1,89 @@
...
+
+crack_data_dict = {
+
+    'external_tight_cracks': {
+        # kg/s/m2 @ 1 Pa
```
chriswmackey, line 13: Do you need to keep repeating this comment?  Can't you just make a note if it in the docstring at the top of the page?

honeybee_energy/ventcool/crack_data.py
```diff
@@ -0,0 +1,89 @@
+# coding':utf-8
+"""Crack data dictionary.
+
+This data is obtained from the DesignBuilder Crack Template. Note that this
+crack data is normalized by area and perimeter for face cracks and subface edges,
+respectively. In EP face cracks are not normalized by area, but subface edges
+are normalized by perimeter.
+"""
+
+crack_data_dict = {
```
chriswmackey, line 10: This should be in all caps `CRACK_DATA_DICT`.  And it should probably something more specific like `CRACK_TEMPLATE_DATA` .

saeranv:
@chriswmackey 

This contains all the changes discussed in the last code review.  Let me know how it looks, thanks.

honeybee_energy/properties/room.py
```diff
@@ -279,6 +286,200 @@ def is_conditioned(self):
...
+        Note that this coefficient is normalized per unit area. The EnergyPlus
+        AirflowNetwork requires an unnormalized value, and thus this value needs to be
+        multiplied by it's corresponding exposed surface area. The normalized area air
+        mass flow coefficient is derived from a zone's infiltration flow rate using the
+        following formula:
+
+            Qva * d = Cqa * dP^n
+
+            where:
+                Cqa: Air mass flow coefficient per unit meter at 1 Pa [kg/m2/s/P^n]
```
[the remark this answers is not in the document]
chriswmackey: Looks good!
image

saeranv:
Thanks @chriswmackey

Just to clarify, are the sphinx errors generated in the command line after I run ``sphinx-build -b html ./docs ./docs/_build/docs``? I must have missed them last time. 

I think I've fixed it, it looks like this now:

image

PR #306, opened 2020-07-30
feat(hvac): Add classes for detailed HVAC templates

chriswmackey:
This commit adds modules and classes for the ~120 HVAC templates that are available in the OpenStudio standards gem.

The classes don't do much computation but they serve the following purposes:

* Organizing/ classifying the 120 templates (grouping them into 16 classes)
* Providing documentation in the form of docstrings
* Handling serialization to dict
* Storing a few optional ECMs like heat recovery and economizers.

They way they are implemented it a bit like the Radiance modifiers in that most of the current functionality is handled by base classes from which everything inherits. But breaking down everything into 16 classes is useful as it means that we can extend the properties of ECMs that are available in the future.

### DIFF
--- a/honeybee_energy/hvac/allair/__init__.py
+++ b/honeybee_energy/hvac/allair/__init__.py
@@ -1,0 +1,20 @@
+"""Template All-air HVAC definitions.
+
+All-air systems provide both ventilation and satisfaction of heating + cooling
+demand with the same stream of warm/cool air. As such, they often have really
+good humidity control. However, because such systems often involve the 
+
+Properties:
+    * HVAC_TYPES_DICT: A dictionary containing pointers to the classes of each
+        HVAC system. The keys of this dictionary are the names of the HVAC
+        classes (eg. 'VAV').
+    * EQUIPMENT_TYPES_DICT: A dictionary containing pointers to the classes of
+        the HVAC systems. The keys of this dictionary are the names of the HVAC
+        systems as they appear in the OpenStudio standards gem and include the
+        specific equipment in the system (eg. 'VAV chiller with gas boiler reheat').
+"""
+from ._base import _AllAirEnumeration
+
+_all_air_types = _AllAirEnumeration(import_modules=True)
+HVAC_TYPES_DICT = _all_air_types.types_dict
+EQUIPMENT_TYPES_DICT = _all_air_types.equipment_types_dict
--- a/honeybee_energy/hvac/allair/_base.py
+++ b/honeybee_energy/hvac/allair/_base.py
@@ -1,0 +1,270 @@
+# coding=utf-8
+"""Base class for All-Air HVAC systems."""
+from __future__ import division
+
+from .._template import _TemplateSystem, _EnumerationBase
+
+from honeybee._lockable import lockable
+from honeybee.typing import valid_string, float_in_range
+from honeybee.altnumber import autosize
+
+import os
+
+
+@lockable
+class _AllAirBase(_TemplateSystem):
+    """Base class for All-Air HVAC systems.
+
+    Args:
+        identifier: Text string for system identifier. Must be < 100 characters
+            and not contain any EnergyPlus special characters. This will be used to
+            identify the object across a model and in the exported IDF.
+        vintage: Text for the vintage of the template system. This will be used
+            to set efficiencies for various pieces of equipment within the system.
+            Choose from the following.
+
+            * DOE Ref Pre-1980
+            * DOE Ref 1980-2004
+            * 90.1-2004
+            * 90.1-2007
+            * 90.1-2010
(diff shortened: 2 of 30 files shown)


honeybee_energy/hvac/allair/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+good humidity control. However, because such systems often involve the 
+
+Properties:
+    * HVAC_TYPES_DICT: A dictionary containing pointers to the classes of each
+        HVAC system. The keys of this dictionary are the names of the HVAC
+        classes (eg. 'VAV').
+    * EQUIPMENT_TYPES_DICT: A dictionary containing pointers to the classes of
+        the HVAC systems. The keys of this dictionary are the names of the HVAC
+        systems as they appear in the OpenStudio standards gem and include the
+        specific equipment in the system (eg. 'VAV chiller with gas boiler reheat').
```
MingboPeng, line 14: I think the term "EQUIPMENT_TYPE" in general might be confusing, as the description here states: it is really a type of systems. Equipment in OpenStudio is more like a part of the system.

it might be better to change it to "SYSTEM_TYPE" or "SYSTEM_TEMPLATE" in general

honeybee_energy/hvac/allair/_base.py
```diff
@@ -0,0 +1,264 @@
...
+            Choose from the following.
+
+            * DOE Ref Pre-1980
+            * DOE Ref 1980-2004
+            * 90.1-2004
+            * 90.1-2007
+            * 90.1-2010
+            * 90.1-2013
+
+        equipment_type: Text for the specific type of the system and equipment
```
MingboPeng, line 32: same "equipment_type" as I mentioned above.

honeybee_energy/hvac/doas/_base.py
```diff
@@ -0,0 +1,213 @@
...
+
+from honeybee._lockable import lockable
+from honeybee.typing import float_in_range
+
+import os
+
+
+@lockable
+class _DOASBase(_TemplateSystem):
+    """Base class for DOAS HVAC systems.
```
MingboPeng, line 15: Base class for HVAC systems with DOAS

honeybee_energy/hvac/heatcool/_base.py
```diff
@@ -0,0 +1,61 @@
...
+from .._template import _TemplateSystem, _EnumerationBase
+
+from honeybee._lockable import lockable
+
+import os
+
+
+@lockable
+class _HeatCoolBase(_TemplateSystem):
+    """Base class for DOAS HVAC systems.
```
MingboPeng, line 14: Base class for heating and cooling systems without ventilation system

MingboPeng:
This will make users to much easier to make ASHRAE baseline system. Thanks @chriswmackey .

Do you know if OpenStudio has any documents for details of each system and case studies?

honeybee_energy/hvac/allair/__init__.py
chriswmackey: I originally called it `system_type` but this was just too similar to the `type` key that all of the schemas have and that refers to the class of the system.  So, if you ask me, "VAV" sounds like a system type but "VAV chiller with gas boiler reheat" sounds like a specification of the equipment within the system. I think it's just better to make the terms different from one another to avoid confusion. We can revisit this again in the honeybee-schema PR if Mostapha feels similarly.

honeybee_energy/hvac/heatcool/_base.py
chriswmackey: Good catch.

chriswmackey:
Hey @MingboPeng ,
I agree that the creation of baseline systems will be a lot easier once we are using the standards gem. So much of the manual process you described to me of setting equipment efficiencies for individual pieces of HVAC equipment can happen in a automated/ streamlined way with these templates.  At the very least, the output here will be a good reference for people setting up baseline Ironbug HVACs such that they can better tweak the properties of individual pieces of equipment.

Unfortunately, the documentation of the standards gem still feels a bit lacking to me but I'll say that it was very helpful for me to know that there are online docs for the standards gem that are generated from the code:
https://www.rubydoc.info/gems/openstudio-standards/
And the Standard class in particular has the methods I am planning to use to set all of these HVAC templates within the honeybee-openstudio-gem:
https://www.rubydoc.info/gems/openstudio-standards/Standard

Beyond that, I get the sense that NREL would point you to [their massive Google Sheets files with all of the individual specifications](https://drive.google.com/drive/folders/1x7yEU4jnKw-gskLBih8IopStwl0KAMEi) or the ASHRAE 90.1 standard itself if you wanted more info beyond the [Create Typical DOE Building measure](https://bcl.nrel.gov/node/85019).

PR #353, opened 2020-09-02
fix(afn): Ensure AirBoundary Faces are properly represented in AFN

chriswmackey:
This commit ensures that the afn.genrate method recognizes Faces with an AirBoundary type and assigns them a Crack object with a very large flow coefficient.  This flow coefficient is derived from the Orifice Equation, using a discharge coefficient of 0.65 for an unobstructed opening.

### DIFF
--- a/honeybee_energy/ventcool/afn.py
+++ b/honeybee_energy/ventcool/afn.py
@@ -2,6 +2,8 @@
 """Functions to generate an Airflow Network for a list of rooms."""
 from __future__ import division
 
+import math
+
 from .crack import AFNCrack
 from .opening import VentilationOpening
 from ._crack_data import CRACK_TEMPLATE_DATA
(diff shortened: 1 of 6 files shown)


honeybee_energy/ventcool/afn.py
```diff
@@ -81,6 +88,13 @@ def _interior_afn(interior_face_groups, int_cracks):
         flow_exp = int_cracks['floorceiling_flow_exp']
         int_floorceiling.properties.energy.vent_crack = AFNCrack(flow_cof, flow_exp)
 
+    # add interior crack leakage for air boundary Faces
+    for int_ab in int_air:
+        # derive (large) flow coefficient from the orifice equation with 0.65 discharge
```
saeranv, line 93: @chriswmackey Here's a suggested revision to the comments to link this flow coefficient derivation back to the ELA formula. It's a more straightforward derivation then using the orifice integration I originally derived the formula from, and so will be more useful as a reference.  

derive (large) flow coefficient for interzone air flow from the formula for mass flow rate from effective leakage area, with a 0.65 discharge.[1]

Note:
[1] LLC, Big Ladder Software. “AirflowNetwork:MultiZone:Surface:EffectiveLeakageArea.” Accessed September 3, 2020. https://bigladdersoftware.com/epx/docs/9-3/input-output-reference/group-airflow-network.html#airflownetworkmultizonesurfaceeffectiveleakagearea.

PR #369, opened 2020-09-10
fix(lighting): Add baseline_watts_per_area to Lighting class

chriswmackey:
This new property will provide a user interface for setting the lighting power density in the baseline model creation workflows.  This property is really only relevant when users are creating their own ProgramTypes from scratch (as opposed to using the ProgramTypes that come with the honeybee-energy-standards package).

The plan is that the baseline model creation workflows will first look for an equivalent ProgramType in the ASHRAE-2004 standard and use the LPD there.  If an equivalent ProgramType in 2004 is not found, this baseline_watts_per_area property will be used instead.  The default value for this property is the ASHRAE 2004 baseline for an office, which is the recommended approach when a given program cannot be classified into any of the categories of the ASHRAE standard.

### DIFF
--- a/honeybee_energy/cli/__init__.py
+++ b/honeybee_energy/cli/__init__.py
@@ -13,6 +13,7 @@ from .translate import translate
 from .settings import settings
 from .simulate import simulate
 from .result import result
+from .baseline import baseline
 
 # command group for all energy extension commands.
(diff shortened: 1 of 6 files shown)


MingboPeng:
I like this approach with new baseline_watts_per_area property, I can imagine that it will be very clear for user to use.

PR #375, opened 2020-09-15
fix(dockerfile): Add honeybee-energy-standards into Dockerfile

chriswmackey:
This commit adds the honeybee-energy-standards package into the docker file as it is required in order to run all of the new baseline model creation commands.

@mostaphaRoudsari  and @AntoineDao , please let me know if I am doing this correctly.  Also, adding in this package will likely increase the size of the docker image by 10 MBs so I am not totally sure this is worth it given that not all recipes will be using the honeybee-energy-standards package.  Let me know if you think we need to set up separate docker images depending on whether the standards package is needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ RUN mkdir -p ladybug_tools/resources/measures/honeybee_openstudio_gem \
 ENV PATH="/home/ladybugbot/.local/bin:${PATH}"
 COPY . honeybee-energy
 RUN pip3 install setuptools wheel\
-    && pip3 install pydantic==1.5.1 honeybee-schema ./honeybee-energy[cli]
+    && pip3 install ./honeybee-energy[cli]
 
 # Set up working directory
 RUN mkdir -p /home/ladybugbot/run/simulation
--- a/honeybee_energy/cli/baseline.py
+++ b/honeybee_energy/cli/baseline.py
@@ -97,7 +97,7 @@ def constructions_2004(model_json, climate_zone, output_file):
     all rooms in the model.
     \n
     Args:
-        model_json: Full path to a Model JSON file.
+        model_json: Full path to a Model JSON file.\n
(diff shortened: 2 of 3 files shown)


AntoineDao (approved):
LGTM. Is this something we imagine would be packaged whenever a user uses the `cli` version of honeybee-energy? If so we might be better served adding the dependency in the `setup.py` file and then running `pip install honeybee-energy[cli]` in the docker container instead.

chriswmackey:
Thanks @AntoineDao .  That's a good point.  The main reason why honeybee-energy-standards isn't packaged with honeybee-energy by default is that honeybee-energy-standards has no Python code in it and it's only a package of 10 MBs of JSONs that can be parsed by honeybee-energy. So I still want to keep open the option of not installing and I don't want to add it to requirements.txt. But adding it to setup.py works.

Along these lines [why do you include honeybee-schema and pydanitc in the docker image here](https://github.com/ladybug-tools/honeybee-energy/blob/master/Dockerfile#L50) when [setup.py already requires honeybee-core[cli]](https://github.com/ladybug-tools/honeybee-energy/blob/master/setup.py#L24) and [honeybee-core[cli] has dependencies on honeybee-schema and pydantic](https://github.com/ladybug-tools/honeybee-core/blob/master/setup.py#L23)?  Can I just delete all of these (seemingly redundant) dependencies and just have them all taken care of by honeybee-energy[cli]?

AntoineDao:
The reason why I did that is unclear, probably because I'm lazy, wanted to be safe and wasn't sure what the dependencies packaged inside the container were 🙃 If we can dedupeclitate these installs by putting them in `setup.py` then I'd say let's do it! 😄

chriswmackey:
Thanks @AntoineDao . I deduplicated everything after googling to see if deduplicate is a word (it is and it's a very cool one!)

I have also been lazy about making some formal documentation of how everything integrates across the dependency tree.  Hopefully, when I get all of this together, it will be clearer when dependencies are duplicated.

PR #495, opened 2020-12-15
fix(docker): Update docker image to OpenStudio 3.1

chriswmackey:
This PR should not be merged until we put the dockerhub credentials into Github so that the CI workflow runs successfully.

CC: @mostaphaRoudsari FYI


mostaphaRoudsari:
This commit should fix the issue: https://github.com/ladybug-tools/honeybee-energy/commit/356454451439c7f4256a1da20a0fd28db78912b4

PR #521, opened 2021-01-06
fix(hvac): Re-make all HVAC  Enumerations

chriswmackey:
This commit also improves the __repr__ for many of the objects to be more human-readable (using display_name instead of identifier) and many of them are now 1 line in order to be more easily printable.


mostaphaRoudsari:
The name of the commit can be misleading! I thought this is for HVAC enumerations and I was pretty excited to see the new human readable ones!

chriswmackey:
That's because it's coming, @mostaphaRoudsari . Sorry, I should have made the PR a draft one until I had added it.

PR #532, opened 2021-01-13
feat(cli): Add  commands to simulate IDF and OSM

chriswmackey:


mostaphaRoudsari:
Nice! Does it mean we will also get new Pollination recipes that support idf and osm directly? 😀

chriswmackey:
Yes it does, @mostaphaRoudsari !

I think I'll do the IDF simulation in a `pollination-honeybee-energy-lite` package that only includes EnergyPlus and not OpenStudio.

PR #611, opened 2021-03-05
feat(docker): improve image build

tymokvo:
# What
- Reduce image size
- Refactor build into its own script for easier local testing

# Why
- To enable faster pulls

# How
- Move blob getting to `build_image.sh`
- Refactor `deploy.sh`
- Base on `python:3.7-slim`

Reduces image size `1.42GB` -> `1.06GB`. Not amazing, but slightly better.

# Testing
I tested this with:
1. `./build_image.sh hbe latest true`
2. `docker run -v $(realpath .)/tests:/home/ladybugbot/run/tests -it -v $(realpath .)/dev-requirements.txt:/home/ladybugbot/run/dev-requirements.txt hbe:latest bash`
3. (inside container now) `pip install -r dev-requirement.txt`
4. `python -m pytest tests/`

All tests should be passing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,41 +4,61 @@ LABEL maintainer="Ladybug Tools" email="info@ladybug.tools"
 
 ARG OPENSTUDIO_VERSION
 ARG OPENSTUDIO_FILENAME
+ARG HONEYBEE_GEM_FILENAME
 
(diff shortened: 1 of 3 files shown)


chriswmackey (approved):
It looks like this should work. I'll aim to test it on pollination before the end of today.

PR #619, opened 2021-03-09
docs(sphinx/conf): support CLI group names not matching CLI file names

santiagogaray:
Hi @chriswmackey, as requested here's is the update on the CLI documentation routine. Will work on the other two repos as you suggested next.


chriswmackey (approved):
LGTM! Thanks!

PR #631, opened 2021-03-20
feat(afn.py): Modify crack logic to match constant infiltration flow.

saeranv:
@chriswmackey 

This PR includes changes to the AFN infiltration calculation, and updated tests to better match infiltration flow through exterior area, including:
1. Incorporate the door and aperture infiltration (from edge cracks) with the area crack flow in their parent face. This is done by adding door and aperture area with parent area when scaling the normalized flow coefficient. 
2. Exposing the delta_pressure parameter for flow coefficient calculations, and modifying docstring to indicate how it should match simulated negative building pressure (relative to EP air node 1 to 2 custom result) to match constant interior flow rate per exterior area.

Note that, removing the edge crack flow (like we discussed in the forums) didn't make a huge difference, but I'm including it here since simplifying the model without having a huge impact on infiltration rate makes this version better then the last.  

I did manage to figure out how to get the AFN infiltration to pretty much exactly match the infiltrate rate per exterior area, but it involves approximating the average pressure difference, (weighted to account for the nonlinear relationship between envelope area and mass flow) from the simulated building, since there was a large difference between the simulated building pressure and our 4 Pa assumption. This was the reason I also exposed the delta_pressure parameter in this PR.  I'll post more details on the forum.

### DIFF
--- a/honeybee_energy/properties/room.py
+++ b/honeybee_energy/properties/room.py
@@ -636,10 +636,23 @@ class RoomEnergyProperties(object):
                 air density at a temperature of 20 C and 101325 Pa).
             delta_pressure: Reference air pressure difference across the building
                 envelope orifice in Pascals used to calculate infiltration crack flow
-                coefficients. If attempting to replicate the room infiltration rate per
-                exterior area, this value should approximate the average positive
-                simulated pressure differences from the exterior to the interior.
-                Default 4 represents typical building pressures.
+                coefficients. The resulting average simulated air pressure difference
+                will roughly equal this delta_pressure times the nth root of the ratio
+                between the simulated and target room infiltration rates::
+
+                    dP_sim = (Q_sim / Q_target)^(1/n) * dP_ref
+
+                    where:
+                        dP: delta_pressure, the reference air pressure difference [Pa]
+                        dP_sim: Simulated air pressure difference [Pa]
+                        Q_sim: Simulated volumetric air flow rate per area [m3/s/m2]
+                        Q_target: Target volumetric air flow rate per area [m3/s/m2]
+                        n: Air mass flow exponent [-]
+
+                If attempting to replicate the room infiltration rate per exterior area,
+                delta_pressure should be set to an approximation of the simulated air
+                pressure difference described in the above formula. Default 4 represents
+                typical building pressures.
         """
         # simplify parameters
         ext_walls, ext_roofs, ext_floors, ext_apertures, ext_doors = exterior_face_groups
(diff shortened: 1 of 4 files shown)


saeranv:
@chriswmackey, I'm moving this to draft in case discussion on forum following my explanation requires changes.

chriswmackey:
Thanks, @saeranv . I'll review shortly.

honeybee_energy/properties/room.py
```diff
@@ -953,22 +966,23 @@ def solve_norm_perimeter_flow_coefficient(norm_area_flow_coefficient, face_area,
...
                 dP: Change in pressure across building envelope [Pa]
                 n: Air mass flow exponent [-]
 
-        Since ``Qv * d / dP^n`` equals ``Cqa`` the normalized area flow coefficient,
+        Since ``(Qva * d) / dP^n`` equals ``Cqa`` the normalized area flow coefficient,
         this can be simplified and rearranged to solve for ``Cql`` with the following
-        formula::
-
-            Cql = Cqa * A / L
+        formula:
```
chriswmackey, line 982: We need the extra line break and the `::` here to make sure that the formula displays correctly in the online docs like so:

https://www.ladybug.tools/honeybee-energy/docs/honeybee_energy.properties.room.html#honeybee_energy.properties.room.RoomEnergyProperties.solve_norm_perimeter_flow_coefficient

chriswmackey (approved):
This is great, @saeranv . Thank you for doing this. I only have one minor comment about one of your docstring formats, which would be nice to fix before we merge. Once we merge this, I'll expose the delta pressure on the HB-Energy components of the Grasshopper plugin.

saeranv:
@chriswmackey 

I made a couple of docstring changes for the sake of accuracy. I also added a description of how the delta_pressure value is related to the simulated delta pressure, for people who are trying to match room infiltration rates with their AFN simulation. I know you try to avoid overly-long docstrings so let me know if you want me to remove it. 
 
I've posted the derivation of that relationship, and justification for exposing the delta_pressure parameter here:
https://discourse.ladybug.tools/t/afn-infiltration-and-ventilation-questions/13471/11?u=saeranvasanthakumar

This out of draft format now, so it's good to go when ready.

chriswmackey (approved):
Thanks @saeranv . This all looks good to me.

To quite PEP8, "long docstrings are good as long as they're maintained." If I have ever been hesitant about them, it's because I've made the mistake of not maintaining my docstrings before. But all of this looks helpful and maintainable.

Merged!

PR #642, opened 2021-03-29
Modify the energy-use-intensity command to accept both folders and files

coditect:
I modified the `energy-use-intensity` command so that it accepts a list of file paths (based on comments by @chriswmackey on ladybug-tools/honeybee-grasshopper-energy#223), folder paths (for backward compatibility), or a mix of the two.


chriswmackey (changes requested):
Thank you, @coditect .  I appreciate you trying to implement this all the right way and I should have given more detail about how the command here should be modified.

If we make this change the way that it is now, it's going to break a lot of other workflows that are using this command.  So we still need the option of passing the `result-folder` but I think it should be just that (an option) rather than an argument.  You can also have an option for `result-paths` where `multiple=True` so that people have the option of either sending a folder or any number of paths to individual sql files.

The command can just raise an assertion error if neither of those options are supplied, and, if both types of arguments are supplied, then the commands just computes EUI across all of them.

I know that it's more work but this will make the command much more reusable across different workflows. Let me know if you have any thoughts or questions and thank you, again.

coditect:
The backwards compatibility you're looking for is already there, but I implemented it using a variadic argument instead of with options as you suggested above.  If you pass a folder to the `result-paths` argument, the modified command will look for `*.sql` files inside that folder just like the current version — I did that specifically so that existing workflows won't have to change.  The `result-paths` argument supports the following inputs:

```sh
# A single folder (just like the current version)
honeybee-energy result energy-use-intensity /path/to/simulation/folder

# Multiple folders
honeybee-energy result energy-use-intensity /path/to/simulation/folder1 /path/to/simulation/folder2 [...]

# A single file
honeybee-energy result energy-use-intensity /path/to/eplusout.sql

# Multiple files
honeybee-energy result energy-use-intensity /path/to/eplusout1.sql /path/to/eplusout2.sql [...]

# A mix of files and folders in any order
honeybee-energy result energy-use-intensity /path/to/simulation/folder /path/to/eplusout.sql [...]
```

Using dedicated options for specifying files and/or folders is certainly possible, it just felt less elegant to me.  Let me know if you still want me to make that change.

chriswmackey (approved):
Ah, I see! I'm sorry, @coditect . I should have looked more carefully at your code and the well-worded docstring that you wrote. Ok, this is all good to go as is.  Thank you, again, and we will continue the conversation over on the honeybee-grasshopper-energy repo.

Merged!

chriswmackey:
Welcome to [the contributors list](https://github.com/ladybug-tools/honeybee-energy/graphs/contributors) @coditect !

PR #652, opened 2021-04-02
fix(sphinx-conf): Fix CLI documentation process

santiagogaray:
Hi @chriswmackey,

Finally some update about your revision. I did some comments at the time in response to your revision in the ``lbt-recipies`` PR but since then I think I made some good progress to handle commands without groups. I am not sure if this will be a bullet-proof solution since there may be some future scenarios that I may have not considered yet. I can discuss this further why future implementations inside ``cli\__init__.py`` may trigger issues, it all depends on how unique these scenarios may be. Otherwise, I tested and it did work on a handful of repos so hopefully this is a good revision to start using.

Also I fixed the group naming by looking at the data inside ``cli\__init__.py``. Feel free to try it in ``honeybee-radiance`` or I can send a PR there as well if you prefer.

Sorry Chris for taking this much time to get this process going, I know you were looking forward to use it in some of your repos. Hope this is almost there if not ready.

Feel free to review when you get a chance.  Hope we can move to other repos if this is good to go!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,6 +14,7 @@
 #
 import os
 import sys
+import re
 import datetime
 now = datetime.datetime.now()
 sys.path.insert(0, os.path.abspath('..'))
@@ -235,8 +236,6 @@ additional documentation reST files(.rst) need to be generated in docs\\cli fold
 
 Note:
     This process assumes that each CLI module represent a group of subcommands.
-    Important: Use hash tables below to address special cases such as
-    command group names different from their module file name.
 """
 # Activate this CLI documentation process
 custom_cli_docs = True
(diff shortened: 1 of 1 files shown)


chriswmackey (approved):
This looks great to me, @santiagogaray . Thank you for getting it together and there is definitely no need to apologize.

Getting these CLI docs right is meant to be an iterative process and we still have plenty of time before the end of the subcontract. The most valuable thing at the end of the day is that we get a clean, reliable solution that we can maintain and apply on other repos into the future. I would be very surprised if we got to that without going back and fixing a few things.

chriswmackey:
The docs are looking great! Thanks again, @santiagogaray!

Issue #207, opened 2020-05-31
Add Internal Mass to honeybee-energy Rooms


chriswmackey:
I know that this feature isn't used as much as some of the other energy simulation features but we should add it for the sake of saying that all legacy features are supported.  I'm sure that it's also useful in making hVAC sizing calculations more realistic.

To do this, we will need to use:
* [InternalMassDefinition](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.0.0-doc/model/html/classopenstudio_1_1model_1_1_internal_mass_definition.html)
* [InternalMass](https://openstudio-sdk-documentation.s3.amazonaws.com/cpp/OpenStudio-3.0.0-doc/model/html/classopenstudio_1_1model_1_1_internal_mass.html)

MaxMarschall:
Hi @chriswmackey, we have a specific project where we will be needing this feature within the next few weeks and it would be great if we could do it in LBT V1 since we also want to use the airflowNetwork. If you are able to bump this one up in the list it would be very much appreciated!

chriswmackey:
This is definitely one of the easier ones to implement as there's not too much to figure out. I'll see if I can get to is soon.

brearleyjonathon:
I am in a similar position to @MaxMarschall ! would be amazing to have in LBT V1! thank you for all your work @chriswmackey

PR #712, opened 2021-06-15
feat(internalmass): Add module and class for InternalMass

chriswmackey:
Resolves https://github.com/ladybug-tools/honeybee-energy/issues/207


brearleyjonathon:
Wow, Chris, thank you! At the risk of showing my ignorance in this powerful world, can you point me towards the resources I would need to implement this update in a similar way that [this](https://grasshopperdocs.com/components/honeybee/addInternalMassToZone.html) legacy component does. I see the internalmass.py in the source code.

PR #752, opened 2021-08-22
Added user_data to honeybee-energy 

TfedUD:
Added user_data to opaque materials and constructions, to facilitate piggy-backing data for hygrothermal simulation and soon (hopefully) also adding some LCA interoperability via EC3 API


honeybee_energy/construction/_base.py
```diff
@@ -154,6 +156,16 @@ def has_shade(self):
...
+    @property
+    def user_data(self):
+        """Get or set an optional dictionary for additional meta data for this object.
+
+        This will be None until it has been set. All keys and values of this
+        dictionary should be of a standard Python type to ensure correct
+        serialization of the object to/from JSON (eg. str, float, int, list, dict)
+        """
+        return self._user_data
+
```
chriswmackey, line 168: You need to include a setter here or else the property will be read-only. Here is an example of a setter in honeybee-core:

https://github.com/ladybug-tools/honeybee-core/blob/master/honeybee/_base.py#L87-L92

honeybee_energy/material/_base.py
```diff
@@ -53,6 +55,16 @@ def display_name(self, value):
...
+    @property
+    def user_data(self):
+        """Get or set an optional dictionary for additional meta data for this object.
+
+        This will be None until it has been set. All keys and values of this
+        dictionary should be of a standard Python type to ensure correct
+        serialization of the object to/from JSON (eg. str, float, int, list, dict)
+        """
+        return self._user_data
+
```
chriswmackey, line 67: We also need a setter here.

honeybee_energy/material/opaque.py
```diff
@@ -330,6 +336,7 @@ def __copy__(self):
             self.specific_heat, self.roughness, self.thermal_absorptance,
             self.solar_absorptance, self._visible_absorptance)
         new_material._display_name = self._display_name
+        new_material._user_data = None if self.user_data is None else self.user_data.copy()
```
chriswmackey, line 339: You will need something like this in the `__copy__` functions of all materials and constructions with user data. Otherwise it will get lost as the object is passed from one Grasshopper component to another.

honeybee_energy/construction/opaque.py
```diff
@@ -200,6 +201,8 @@ def from_dict(cls, data):
         new_obj = cls(data['identifier'], mat_layers)
         if 'display_name' in data and data['display_name'] is not None:
             new_obj.display_name = data['display_name']
+        if 'user_data' in data and data['user_data'] is not None:
+            opaqueConstruction.user_data = data['user_data']
```
chriswmackey, line 205: You also need to add a couple of lines like this into the `from_dict_abridged` classmethod.

chriswmackey (changes requested):
Thank you for sending the PR, @TfedUD . I agree that this will be very useful and it is certainly a part of the long-term vision.

It is a pretty big undertaking, though, since we probably should not merge a `user_data` PR into the official codebase until we can say broadly that "honeybee-energy supports `user_data`." It's just been extremely useful to explain to people that "honeybee-core supports `user_data` but honeybee-energy does not" and I wouldn't want that explanation to change to "some of honeybee-energy has `user_data` and it's unclear which parts lack it." If you are up for the challenge, though, I think we would need it implemented on the following objects in order to be able to merge:

* All constructions
* All materials
* The two types of schedules (Ruleset and FixedInterval)
* PorgramTypes
* ConstructionSets
* All HVAC systems
* InternalMass objects

I think we probably don't need it on `ScheduleTypeLimits` or the individual Load objects (including all of the ones associated with natural ventilation, daylight controls, etc.). These ones almost never appear without a parent object and so the parent can probably carry whatever data these sub-objects would have had. Maybe we might want it on SimulationParameter but I don't think we should do it in the first merge and we can wait and see what people think on that one.

I left some things to be corrected in your current implementation and I would also strongly recommend that you add some tests. Adding a few lines of code to existing unit tests is probably fine for this case and you don't have to add brand new test functions. However, I would recommend at least making sure that the `user_data` survives the `duplicate()` process (which calls `__copy__` under the hood) and that is also survives the `to_dict()` and then `from_dict()` translation process. I can see already that the `user_data` on the OpaqueConstruction class would not survive duplication.

If you manage to get a corrected version implemented on all of those objects, then we can merge and add the new `user_data` properties into [honeybee-schema](https://github.com/ladybug-tools/honeybee-schema), which will allow all of the authors of Ladybug Tools plugins to make use of it. 

At any point, you should feel free to re-request my review to look over any updates. Let us know what you think!

TfedUD:
@chriswmackey   Awesome!   Will do for sure! thank you!

TfedUD:
@chriswmackey  re: [This HVAC push](https://github.com/ladybug-tools/honeybee-energy/pull/752/commits/8efd3c0ddaf10c12498a1c7e97d8667b2eb45fa4)

This is more so a py dev in general question tho the context makes it make sense to me:

Adding ```user_data``` following the mats/constrs flow/lessons learned: 
  _template has the 'last' from/to_dict and copy (sans idealair.py)
  
  With allair, doas etc classes inheriting _TemplateSystem, those are not needing any attention as the __copy__ in the inherited class covers the ```user_data```?
  
  Or does up to each 'doas / allair / heatcool'  level need to be attended to?
  ```python
  class _DOASBase(_TemplateSystem): # need to additionally get the user_data love to the to/from and copy?

TfedUD:
@chriswmackey  Finally some time to get into the tests:  I've implemented on one existing test:
```python
def test_opaque_dict_methods():
    """Test the to/from dict methods."""
    concrete = EnergyMaterial('Concrete', 0.15, 2.31, 2322, 832)
    insulation = EnergyMaterial('Insulation', 0.05, 0.049, 265, 836)
    wall_gap = EnergyMaterial('Wall Air Gap', 0.1, 0.67, 1.2925, 1006.1)
    gypsum = EnergyMaterial('Gypsum', 0.0127, 0.16, 784.9, 830)
    userdatadict = {"dictKey":"KeyName",
                    "KeyWord":"KeyValue"}
    wall_constr = OpaqueConstruction(
        'Wall Construction', [concrete, insulation, wall_gap, gypsum])
    wall_constr.user_data = userdatadict
    constr_dict = wall_constr.to_dict()
    new_constr = OpaqueConstruction.from_dict(constr_dict)
    assert constr_dict == new_constr.to_dict()
    assert wall_constr.user_data == new_constr.user_data
```

TfedUD:
When changing: 
```assert wall_constr.user_data == new_constr.user_data``` 
     to '!=':
    The following assertion error is raised via pytest, is this sufficient to apply to the rest of the objects in the tests, or is there anything additional you would want implemented in the tests?
***
```
E       AssertionError: assert {'KeyWord': 'KeyValue', 'dictKey': 'KeyName'} != {'KeyWord': 'KeyValue', 'dictKey': 'KeyName'}E        +  where {'KeyWord': 'KeyValue', 'dictKey': 'KeyName'} = Construction,\n Wall Construction,        !- name\n Concrete,                 !- layer 1\n Insulation,               !- layer 2\n Wall Air Gap,             !- layer 3\n Gypsum;                   !- layer 4.user_dataE        +  and   {'KeyWord': 'KeyValue', 'dictKey': 'KeyName'} = Construction,\n Wall Construction,        !- name\n Concrete,                 !- layer 1\n Insulation,               !- layer 2\n Wall Air Gap,             !- layer 3\n Gypsum;                   !- layer 4.user_data
```

**Comment by TfedUD** [comment]
Have updated userdatadict to an 'all the types' dict per reccomendation of a [mentor](https://discourse.ladybug.tools/u/edpmay/summary) in work pytest error fixing

**Comment by TfedUD** [comment]
Forgot to check a couple things

**Comment by TfedUD** [comment]
if there was multiple notifications: I apologize

**Comment by TfedUD** [comment]
actually: I've ID'd a few issues while using this in the dev of a Passive House proj I'm colabing on: will fix shortly

File: honeybee_energy/hvac/allair/_base.py [hunk 5]
```diff
@@ -63,12 +63,13 @@ class _AllAirBase(_TemplateSystem):
...
         * schedules
+        * user_data
     """
     __slots__ = ('_economizer_type', '_sensible_heat_recovery', '_latent_heat_recovery',
-                 '_demand_controlled_ventilation')
+                 '_demand_controlled_ventilation', '_user_data')
     ECONOMIZER_TYPES = ('NoEconomizer', 'DifferentialDryBulb', 'DifferentialEnthalpy',
                         'DifferentialDryBulbAndEnthalpy', 'FixedDryBulb',
-                        'FixedEnthalpy', 'ElectronicEnthalpy')
+                        'FixedEnthalpy', 'ElectronicEnthalpy', '_user_data')
```
chriswmackey, line 72: This one looks like you copy/pasted things into the wrong tuple. Just remove this one and it should be good.

honeybee_energy/hvac/allair/_base.py
```diff
@@ -63,12 +63,13 @@ class _AllAirBase(_TemplateSystem):
         * latent_heat_recovery
         * demand_controlled_ventilation
         * schedules
+        * user_data
     """
     __slots__ = ('_economizer_type', '_sensible_heat_recovery', '_latent_heat_recovery',
-                 '_demand_controlled_ventilation')
+                 '_demand_controlled_ventilation', '_user_data')
```
chriswmackey, line 69: This is not the correct way to use `__slots__`. If the variable is already defined in the class that you are inheriting from, you should not repeat it here. Otherwise you are un-doing a lot of the memory saving that we get from using `__slots__`. So you should remove this from here and from the `__slots___` of all other HVAC system classes. More info is here:

https://stackoverflow.com/questions/1816483/how-does-inheritance-of-slots-in-subclasses-actually-work

honeybee_energy/constructionset.py
```diff
@@ -1314,21 +1390,23 @@ class DoorConstructionSet(object):
...
     """
     __slots__ = ('_exterior_construction', '_interior_construction',
                  '_exterior_glass_construction', '_interior_glass_construction',
-                 '_overhead_construction', '_locked')
+                 '_overhead_construction', '_locked', '_user_data')
 
     def __init__(self, exterior_construction=None, interior_construction=None,
                  exterior_glass_construction=None, interior_glass_construction=None,
-                 overhead_construction=None):
+                 overhead_construction=None, _user_data=None):
```
chriswmackey, line 1401: The `_user_data=None` should be removed from the `__init__` arguments. This keeps things consistent with the other classes and you aren't using this argument now.

honeybee_energy/hvac/doas/_base.py
```diff
@@ -58,9 +58,10 @@ class _DOASBase(_TemplateSystem):
         * demand_controlled_ventilation
         * doas_availability_schedule
         * schedules
+        * user_data
     """
     __slots__ = ('_sensible_heat_recovery', '_latent_heat_recovery',
-                 '_demand_controlled_ventilation', '_doas_availability_schedule')
+                 '_demand_controlled_ventilation', '_doas_availability_schedule', '_user_data')
```
chriswmackey, line 64: Remove from `__slots__` since it's already in the parent class.

honeybee_energy/hvac/idealair.py
```diff
@@ -77,7 +77,7 @@ class IdealAirSystem(_HVACSystem):
                  '_sensible_heat_recovery', '_latent_heat_recovery',
                  '_heating_air_temperature', '_cooling_air_temperature',
                  '_heating_limit', '_cooling_limit', '_heating_availability',
-                 '_cooling_availability')
+                 '_cooling_availability', '_user_data',)
```
chriswmackey, line 80: Remove from `__slots__` since it's already in the parent class.

honeybee_energy/load/infiltration.py
```diff
@@ -57,9 +57,10 @@ class Infiltration(_LoadBase):
         * constant_coefficient
         * temperature_coefficient
         * velocity_coefficient
+        * user_data
     """
     __slots__ = ('_flow_per_exterior_area', '_schedule', '_constant_coefficient',
-                 '_temperature_coefficient', '_velocity_coefficient')
+                 '_temperature_coefficient', '_velocity_coefficient', '_user_data')
```
chriswmackey, line 63: Remove from `__slots__` since it's already in the parent class.

honeybee_energy/load/lighting.py
```diff
@@ -46,9 +46,10 @@ class Lighting(_LoadBase):
         * visible_fraction
         * convected_fraction
         * baseline_watts_per_area
+        * user_data
     """
     __slots__ = ('_watts_per_area', '_schedule', '_return_air_fraction',
-                 '_radiant_fraction', '_visible_fraction', '_baseline_watts_per_area')
+                 '_radiant_fraction', '_visible_fraction', '_baseline_watts_per_area', '_user_data')
```
chriswmackey, line 52: Remove from `__slots__` since it's already in the parent class.

honeybee_energy/load/lighting.py
```diff
@@ -63,6 +64,7 @@ def __init__(self, identifier, watts_per_area, schedule, return_air_fraction=0.0
         self.radiant_fraction = radiant_fraction
         self.visible_fraction = visible_fraction
         self.baseline_watts_per_area = None  # can be set by the user later
+        self._user_data = None
```
chriswmackey, line 67: There's no need to do this here since we are already initializing `_LoadBase`, which will set this variable to `None`. I think you'll also have to remove this for a lot of other Load classes.

honeybee_energy/load/people.py
```diff
@@ -53,9 +53,10 @@ class People(_LoadBase):
         * activity_schedule
         * radiant_fraction
         * latent_fraction
+        * user_data
     """
     __slots__ = ('_people_per_area', '_occupancy_schedule', '_activity_schedule',
-                 '_radiant_fraction', '_latent_fraction')
+                 '_radiant_fraction', '_latent_fraction', '_user_data')
```
chriswmackey, line 59: Remove from `__slots__` since it's already in the parent class.

honeybee_energy/load/setpoint.py
```diff
@@ -51,7 +51,7 @@ class Setpoint(_LoadBase):
         * dehumidifying_setback
     """
     __slots__ = ('_heating_schedule', '_cooling_schedule', '_humidifying_schedule',
-                 '_dehumidifying_schedule')
+                 '_dehumidifying_schedule', '_user_data')
```
chriswmackey, line 54: Remove from `__slots__` since it's already in the parent class.

honeybee_energy/load/ventilation.py
```diff
@@ -63,7 +63,7 @@ class Ventilation(_LoadBase):
         * schedule
     """
     __slots__ = ('_flow_per_person', '_flow_per_area', '_flow_per_zone',
-                 '_air_changes_per_hour', '_schedule')
+                 '_air_changes_per_hour', '_schedule', '_user_data')
```
chriswmackey, line 66: Remove from `__slots__` since it's already in the parent class.

honeybee_energy/load/ventilation.py
```diff
@@ -74,6 +74,7 @@ def __init__(self, identifier, flow_per_person=0, flow_per_area=0, flow_per_zone
         self.flow_per_zone = flow_per_zone
         self.air_changes_per_hour = air_changes_per_hour
         self.schedule = schedule
+        self._user_data = None
```
chriswmackey, line 77: Remove this line since it already happens when initializing the parent class.

honeybee_energy/load/setpoint.py
```diff
@@ -68,6 +68,7 @@ def __init__(self, identifier, heating_schedule, cooling_schedule,
         self.cooling_schedule = cooling_schedule
         self.humidifying_schedule = humidifying_schedule
         self.dehumidifying_schedule = dehumidifying_schedule
+        self._user_data = None
```
chriswmackey, line 71: Remove this line since it already happens when initializing the parent class.

honeybee_energy/load/people.py
```diff
@@ -67,6 +68,7 @@ def __init__(self, identifier, people_per_area, occupancy_schedule,
         self.activity_schedule = activity_schedule
         self.radiant_fraction = radiant_fraction
         self.latent_fraction = latent_fraction
+        self._user_data = None
```
chriswmackey, line 71: Remove this line since it already happens when initializing the parent class.

honeybee_energy/load/infiltration.py
```diff
@@ -71,6 +72,7 @@ def __init__(self, identifier, flow_per_exterior_area, schedule,
         self.constant_coefficient = constant_coefficient
         self.temperature_coefficient = temperature_coefficient
         self.velocity_coefficient = velocity_coefficient
+        self._user_data = None
```
chriswmackey, line 75: Remove this line since it already happens when initializing the parent class.

honeybee_energy/hvac/allair/_base.py
```diff
@@ -84,6 +85,7 @@ def __init__(self, identifier, vintage='ASHRAE_2019', equipment_type=None,
         self.sensible_heat_recovery = sensible_heat_recovery
         self.latent_heat_recovery = latent_heat_recovery
         self.demand_controlled_ventilation = demand_controlled_ventilation
+        self._user_data = None
```
chriswmackey, line 88: Remove this line since it already happens when initializing the parent class.

honeybee_energy/material/_base.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, identifier):
         self._locked = False
         self.identifier = identifier
         self._display_name = None
+
```
chriswmackey, line 29: It looks like you accidentally added an extra blank line here.

honeybee_energy/material/gas.py
```diff
@@ -34,7 +34,7 @@ class _EnergyWindowMaterialGasBase(_EnergyMaterialWindowBase):
                           'Xenon': (158.33970642, 0.0, 0.0)}
     MOLECULARWEIGHTS = {'Air': 28.97, 'Argon': 39.948,
                         'Krypton': 83.8, 'Xenon': 131.3}
-    __slots__ = ('_thickness',)
+    __slots__ = ('_thickness')
```
chriswmackey, line 37: This is going to break a lot of things since you turned the tuple into a single string. Please put the comma back before we merge.

honeybee_energy/material/gas.py
```diff
@@ -332,13 +332,15 @@ class EnergyWindowMaterialGas(_EnergyWindowMaterialGasBase):
         * specific_heat
         * density
         * prandtl
+        * user_data
     """
-    __slots__ = ('_gas_type',)
+    __slots__ = ('_gas_type','_user_data')
```
chriswmackey, line 337: Why are you adding `user_data` separately to each and every material class. Just implement it in `_EnergyMaterialBase` here:
https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/material/_base.py#L10
Then you don't need to specify it in all of these separate `__slots__` or use duplicated `@properties`. This also means that you won't need to set `self._user_data = None` in all of the `__init__` since it will all be handled in `_EnergyMaterialBase.__init__`.

tests/conftest.py
```diff
@@ -0,0 +1,16 @@
+import pytest
```
chriswmackey, line 1: This is a good way to structure the tests but I would usually call something like this `userdatadict` a "Fixture". So can we just rename this module to `fixutres/userdata.py`? That will be a lot clearer and more specific than `conftest.py`.

chriswmackey (commented):
Thanks, @TfedUD .

Things are generally looking good and I see that you've implemented it on all of the classes that I listed. Good job!

The only case that really needs to be refactored is the Material classes. You should just move the `@property`, the `__slots__`, and the `self._user_data = None` in `__init__` to be on the [_EnergMaterialBase](https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/material/_base.py#L10 class.  Oh, and you introduced a bug into the Gas material classes by chainging the `__slots__` from a tuple to a string (deleting a necessary comma).

There are a few other cases where you have unnecessary lines of code because things are already taken care of with inheritance. I marked all of those with a comment.

Once you remove the unnecessary lines of code and refactor the edits you made to the material classes, then you can request my review one more time and we'll merge if all looks good.

TfedUD:
@chriswmackey  Awesome thank you. this has been a great learning/doing experience; I should have all the aforemention resolved today

honeybee_energy/hvac/allair/_base.py
TfedUD: Oops, see commit: "wrong touple"

honeybee_energy/hvac/allair/_base.py
TfedUD: AHHHHH Okay cool thaat made sense of ```__slots__```  ty!! commit: allair slots edit

honeybee_energy/constructionset.py
TfedUD: commit: constructionSet init args

Yeahh.. I'm not sure how I ended up doing that for some reason lol

honeybee_energy/hvac/doas/_base.py
TfedUD: commit: doas /idealair slots edit

honeybee_energy/hvac/idealair.py
TfedUD: commit: doas /idealair slots edit

honeybee_energy/material/gas.py
TfedUD: ooops grossly overlooked that.
fixed, commit: fixed accidental comma removal

honeybee_energy/load/infiltration.py
TfedUD: commit: loads cleaned up

honeybee_energy/load/lighting.py
TfedUD: commit: loads cleaned up

honeybee_energy/load/lighting.py
TfedUD: commit: loads cleaned up

honeybee_energy/load/setpoint.py
TfedUD: commit: loads cleaned up

honeybee_energy/load/ventilation.py
TfedUD: commit: loads cleaned up

honeybee_energy/load/ventilation.py
TfedUD: commit: loads cleaned up

honeybee_energy/hvac/allair/_base.py
TfedUD: yeah... slots: this is my first time writing using slots: apologies

honeybee_energy/material/_base.py
TfedUD: both refactored and  adressed this additional line
commit:  refatcored materials: I now understand __slots__

tests/conftest.py
TfedUD: yeahh.... my bad.. I go 10000km/hr at all times: adhd needing intensity and all: I meant to circle back on that but it got lost in the excitement

TfedUD:
@chriswmackey 
All but the re-formatting/better dir org has been adressed: will finish tomorrow

TfedUD:
All test pass. about to close up shop at day job: will do a final double check when back home later today

honeybee_energy/material/gas.py
```diff
@@ -332,8 +332,9 @@ class EnergyWindowMaterialGas(_EnergyWindowMaterialGasBase):
         * specific_heat
         * density
         * prandtl
+        * user_data
     """
-    __slots__ = ('_gas_type',)
+    __slots__ = ('_gas_type')
```
chriswmackey, line 337: This is still a bug but I'll fix it after we merge.

honeybee_energy/material/glazing.py
```diff
@@ -102,7 +103,7 @@ def __init__(self, identifier, thickness=0.003, solar_transmittance=0.85,
         self._solar_reflectance_back = None
         self._visible_reflectance = 0
         self._visible_reflectance_back = None
-
+        self._user_data = None
```
chriswmackey, line 106: This is unnecessary but I'll fix it after we merge

TfedUD:
Hi!!  so everything looks good: Cleaned everything up, applied the fixture from a fixture dir.
all tests pass on my machine for HBE buut:

the "some checks were not successful" in the automated action (cool af btw)

    From looking through the details: it looks like issues on the HB side not the HBE side so I am assuming it is schema related? as the new additions to HBE have not been added to the HB schema?

chriswmackey (approved):
Thanks for all of the updates, @TfedUD . It looks like you addressed almost all comments and it is good enough to merge now. There were a couple of lines that still have to be tweaked but I want to merge in this PR and then do some conflict resolution with [this PR](https://github.com/ladybug-tools/honeybee-energy/pull/757) now. So I am just going to merge this now and I'll take care of those two small tweaks as I am doing conflict resolution.

Thanks again for getting this all together. I know that a lot of people will be looking forward to what they can do with these new `user_data` keys.

honeybee_energy/material/glazing.py
TfedUD: ARG how did I miss this I just looked through everything!

TfedUD:
TY! been great fun!

chriswmackey:
FYI, the reason why the tests didn't pass is that one of the tests has a dependency on another URL and so it fails sometimes on the CI. I should probably write a better unit test at some point but you should see that the test should pass after I just triggered a re-run.

TfedUD:
image

chriswmackey:
All released and it's official!
And I patched the other two issues [here](https://github.com/ladybug-tools/honeybee-energy/pull/758).

Thanks again for the big effort here and welcome to the [contributor's list](https://github.com/ladybug-tools/honeybee-energy/graphs/contributors), @TfedUD !

TfedUD:
AHHHH!! No problem!! LBT's literally why I started writing code! literally thrilled to have been able to contribute!

PR #799, opened 2021-12-11
Added greenroof material

TfedUD:
Added greenroof material:
  All tests pass except:
      ```def test_greenroof_init_from_idf():```
```
    def parse_idf_string(idf_string, expected_type=None):
        """Parse an EnergyPlus string of a single object into a list of values.

        Args:
            idf_string: An IDF string for a single EnergyPlus object.
            expected_type: Text representing the expected start of the IDF object.
                (ie. WindowMaterial:Glazing). If None, no type check will be performed.

        Returns:
            ep_fields -- A list of strings with each item in the list as a separate field.
            Note that this list does NOT include the string for the start of the IDF
            object. (ie. WindowMaterial:Glazing)
        """
        idf_string = idf_string.strip()
        if expected_type is not None:
>           assert idf_string.startswith(expected_type), 'Expected EnergyPlus {} ' \
                'but received a different object: {}'.format(expected_type, idf_string)
E           AssertionError: Expected EnergyPlus Material, but received a different object: Material:RoofVegetation,
E           myroof,                   !- name
E           0.5,                      !- height of plants
E           2.0,                      !- leaf area index
E           0.25,                     !- leaf reflectivity
E           0.90,                     !- leaf emissivity
E           350.0,                    !- minimum stomatal resistance
E           GreenRoofDIRT,            !- soil layer name
E           MediumRough,              !- roughness
E           0.5,                      !- thickness
E           0.45,                     !- conductivity of dry soil
E           1250.0,                   !- density of dry soil
E           950.0,                    !- specific heat of dry soil
E           0.74,                      !- thermal absorptance
E           0.6,                      !- solar absorptance
E           0.5,                      !- visible absorptance
E           0.4,                      !- saturation volumetric moisture content of the soil layer
E           0.9,                     !- residual volumetric moisture content of the soil layer
E           0.4,                      !- initial volumetric moisture content of the soil layer
E           Simple;                   !- moisture diffusion calculation method

honeybee_energy\reader.py:20: AssertionError
```
```
    @classmethod
    def from_idf(cls, idf_string):
        """Create an EnergyMaterial from an EnergyPlus text string.

        Args:
            idf_string: A text string fully describing an EnergyPlus material.
        """
        ep_strs = parse_idf_string(idf_string, 'Material:RoofVegetation,')
        return cls(*ep_strs)
```
not sure why it's expecting material when being given roof mat

### DIFF
--- a/honeybee_energy/material/opaque.py
+++ b/honeybee_energy/material/opaque.py
@@ -871,6 +871,13 @@ class EnergyMaterialGreenRoof(_EnergyMaterialOpaqueBase):
 
     @classmethod
     def from_idf(cls, idf_string):
+        """Create an EnergyMaterial from an EnergyPlus text string.
+
+        Args:
+            idf_string: A text string fully describing an EnergyPlus material.
+        """
+        ep_strs = parse_idf_string(idf_string, 'Material:RoofVegetation,')
+
         pass
 
     @classmethod
@@ -928,8 +935,8 @@ class EnergyMaterialGreenRoof(_EnergyMaterialOpaqueBase):
             in data and data['res_vol_moist_cont'] is not None else 0.01
         ivmc = data['init_vol_moist_cont'] if 'init_vol_moist_cont' \
             in data and data['init_vol_moist_cont'] is not None else 0.1
-        mdc = data['moisture_dif_calculation'] if 'moisture_dif_calculation' in data and \
-            data['moisture_dif_calculation'] is not None else 'Simple'
+        mdc = data['moist_dif_calc'] if 'moist_dif_calc' in data and \
+            data['moist_dif_calc'] is not None else 'Simple'
 
         new_mat = cls(rough, thick, t_abs, s_abs, v_abs, dense, sp_ht, p_h, lai,
                       l_r, l_e, msr, sln, svmc, rvmc, ivmc, mdc)
@@ -942,4 +949,19 @@ class EnergyMaterialGreenRoof(_EnergyMaterialOpaqueBase):
 
     def to_idf(self):
         """Get an EnergyPlus string representation of the material."""
-        pass
+        values = (self.identifier, self.plant_height, self.leaf_area_ind,
+                  self.leaf_reflectivity, self.leaf_emissivity, self.min_stomatal_res,
+                  self.soil_layer_name, self.roughness, self.thickness, self.conductivity,
+                  self.density, self.specific_heat, self.thermal_absorptance,
+                  self.solar_absorptance, self.visible_absorptance,
+                  self.sat_vol_moist_cont, self.res_vol_moist_cont, self.init_vol_moist_cont,
+                  self.moist_dif_calc)
+        comments = ('name', 'height of plants', 'leaf area index', 'leaf reflectivity',
+                    'leaf emissivity', 'minimum stomatal resistance', 'soil layer name',
+                    'roughness', 'thickness', 'conductivity of dry soil', 'density of dry soil',
+                    'specific heat of dry soil', 'thermal absorptance', 'solar absorptance',
+                    'visible absorptance', 'saturation volumetric moisture content of the soil layer',
+                    'residual volumetric moisture content of the soil layer',
+                    'initial volumetric moisture content of the soil layer',
+                    'moisture diffusion calculation method')
+        return generate_idf_string('Material:RoofVegetation', values, comments)
--- a/honeybee_energy/material/opaque.py
+++ b/honeybee_energy/material/opaque.py
@@ -653,10 +653,10 @@ class EnergyMaterialGreenRoof(_EnergyMaterialOpaqueBase):
     """Green Roof Material
     https://bigladdersoftware.com/epx/docs/9-5/input-output-reference/group-surface-construction-elements.html#materialroofvegetation
     """
-    __slots__ = ('_plant_height', '_leaf_area_ind', '_leaf_reflectivity',
+    __slots__ = ('_thickness', '_conductivity', '_density', '_specific_heat', '_roughness',
+                 '_thermal_absorptance', '_solar_absorptance', '_visible_absorptance',
+                 '_plant_height', '_leaf_area_ind', '_leaf_reflectivity',
                  '_leaf_emissivity', '_min_stomatal_res', '_soil_layer_name',
-                 '_roughness', '_thickness', '_conductivity', '_density',
-                 '_specific_heat', '_thermal_absorptance', '_solar_absorptance', '_visible_absorptance',
                  '_sat_vol_moist_cont', '_res_vol_moist_cont', '_init_vol_moist_cont',
                  '_moist_dif_calc')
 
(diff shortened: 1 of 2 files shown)


honeybee_energy/material/opaque.py
```diff
@@ -638,3 +646,477 @@ def __copy__(self):
...
+            self._moist_dif_calc = 'Simple'
+
+    @classmethod
+    def from_idf(cls, idf_string):
+        """Create an EnergyMaterial from an EnergyPlus text string.
+
+        Args:
+            idf_string: A text string fully describing an EnergyPlus material.
+        """
+        ep_strs = parse_idf_string(idf_string, "Material:RoofVegetation, ")
```
mostaphaRoudsari, line 958: Hi @TfedUD! The issue that you are facing might be because of the extra space here which makes the assertion fail. Try this change:

```suggestion
        ep_strs = parse_idf_string(idf_string, "Material:RoofVegetation,")
```

TfedUD: I had EnergyMaterial.from_idf in tests rather than EnergyMaterialGreenRoof.from_idf lol.. arg.

now that runs since I fixed: need to tweak from_idf method a wee bit:
But should be good shortly

TfedUD:
@mostaphaRoudsari good to go!

mostaphaRoudsari:
Thanks! I assigned @chriswmackey to review and comment.

TfedUD:
:metal: WindowProperty:FrameAndDivider time!

TfedUD:
oooof:
 Chris when you see this:
    I'm using autopep8: It looks like it reformatted a bunch code I didn't add

mostaphaRoudsari:
@TfedUD, you should set the max line length to 89 in the PEP8 settings to avoid those. That's the number that we use for the ladybug tools project.

TfedUD:
> @TfedUD, you should set the max line length to 89 in the PEP8 settings to avoid those. That's the number that we use for the ladybug tools project.

Ya know: one of these days I'm going to remember that I know how to computer and will problem solve accordingly :sweat_smile:

honeybee_energy/material/opaque.py
```diff
@@ -638,3 +646,480 @@ def __copy__(self):
...
+            This parameter is used when calculating the long wavelength radiant exchange 
+            at the leaf surfaces. Values for this field must be between 
+            0.8 and 1.0 (with 1.0 representing “black body” conditions). Default is .95.
+        min_stomatal_res: This field represents the resistance of the plants to moisture 
+            transport. It has units of s/m. Plants with low values of stomatal resistance 
+            will result in higher evapotranspiration rates than plants with high resistance. 
+            Values for this field must be in the range of 50.0 to 300.0. Default is 180.
+        soil_layer_name: This field is a unique reference name that the user assigns to the 
+            soil layer for a particular ecoroof. This name can then be referred to by 
+            other input data. Default is Green Roof Soil.
```
chriswmackey, line 698: I think we should not expose this property since it is not one that people can use from Honeybee.

In order to keep OpaqueConstructions as self-contained objects, we have users specify explicit EnergyMaterial layers for the construction and they are not allowed to pass the name/identifier of the layer like "Green Roof Soil". So I recommend removing this `soil_layer_name` as a set-able property from this class and we'll just always assume the EnergyPlus default value of "Green Roof Soil" when we serialize the material to and from IDF.

If you want to enable an EnergyPlus-type workflow for the soil layer in Honeybee, you can still have a read-only property on this class called `soil_layer`. But it should return an `EnergyMaterial` instance that has the same thickness, conductivity, etc. as the green roof material. You can generate it on the fly when the property is called such that, if the green roof material properties change, so do the soil layer properties. The identifier of this soil material should probably be derived from the `EnergyMaterialGreenRoof` identifier so that it doesn't conflict as easily with other possible green roof materials in the model.

This way, if people want to make a version of the construction that only has the soil without vegetation, they can easily do so.

honeybee_energy/material/opaque.py
```diff
@@ -638,3 +646,480 @@ def __copy__(self):
...
+        sat_vol_moist_cont: The field allows for user input of the saturation moisture 
+            content of the soil layer. Maximum moisture content is typically less than 
+            0.5. Range is [.1,.5] with the default being .3.
+        res_vol_moist_cont: The field allows for user input of the residual moisture 
+            content of the soil layer. Default is .01, range is [.01,.1].
+        init_vol_moist_cont: The field allows for user input of the initial moisture 
+            content of the soil layer. Range is (.05, .5] with the default being .1.
+        moist_dif_calc: The field allows for two models to be selected: Simple or Advanced. 
+            EnergyPlus Currently supports only the Simple Moisture Diffusion Calculation Method.
```
chriswmackey, line 707: I see that you copy/pasted this from the EnergyPlus Input/Output reference but this description very clearly seems to be out of date since the Advanced method clearly can be used. So you should delete this last sentence and probably replace it with a sentence saying that the timestep of the EnergyPlus simulation should be increased to 20 when using the Advanced model. I agree that the default should definitely be "Simple."

honeybee_energy/material/opaque.py
```diff
@@ -638,3 +646,480 @@ def __copy__(self):
...
+        soil_layer_name: This field is a unique reference name that the user assigns to the 
+            soil layer for a particular ecoroof. This name can then be referred to by 
+            other input data. Default is Green Roof Soil.
+        sat_vol_moist_cont: The field allows for user input of the saturation moisture 
+            content of the soil layer. Maximum moisture content is typically less than 
+            0.5. Range is [.1,.5] with the default being .3.
+        res_vol_moist_cont: The field allows for user input of the residual moisture 
+            content of the soil layer. Default is .01, range is [.01,.1].
+        init_vol_moist_cont: The field allows for user input of the initial moisture 
+            content of the soil layer. Range is (.05, .5] with the default being .1.
```
chriswmackey, line 705: the `( ]` seem to be a typo

chriswmackey: Also, it seems that all of these values are fractional (without units) and it would be good to clarify this.

honeybee_energy/material/opaque.py
```diff
@@ -638,3 +646,480 @@ def __copy__(self):
...
+    def moist_dif_calc(self):
+        return(self._moist_dif_calc)
+
+    @moist_dif_calc.setter
+    def moist_dif_calc(self, mdc):
+        if mdc == 'Simple':
+            self._moist_dif_calc = mdc
+        else:
+            self._moist_dif_calc = 'Simple'
+
```
chriswmackey, line 950: Can you add two read-only properties for `r_value` and `u_value`? You can use the same exact method that you see for the `EnergyMaterial` class where they are derived from the soil thickness and conductivity. I just know that we'll need these values in order to compute OpaqueConstruction R-values that contain this green roof material.

While you are at it , it would also be nice to have read-only properties for `mass_area_density` and `area_heat_capacity` (using the same methods you see on `EnergyMaterial`).  We're not using them now but they will probably be useful at some point for evaluating thermal mass and time constants.

chriswmackey: And one more request: in order for this material to be usable in workflows involving coordination between Radiance and EnergyPlus (eg. the comfort maps), this class needs to have methods for `to_radiance_solar` and `to_radiance_visible`. Can you add these methods using the same code as the `EnergyMaterial` class, where you return a plastic material with the correct reflectance?

chriswmackey (commented):
Hi @TfedUD ,

Thanks for getting this together and it is generally looking good. My only major comments are that we should get rid of the `soil_layer_name` property (since we can't use it) and there are a few other read-only properties and methods that have to bee added in order for the class to play nicely with the other modules.

Once these two are implemented, we should be good to merge.

honeybee_energy/material/opaque.py
TfedUD: Oooof...... did I read this wrong or does someone need to email Aaron and tell him the docs are out of date?
[E+9.6 I/O link to highlight](https://bigladdersoftware.com/epx/docs/9-6/input-output-reference/group-surface-construction-elements.html#:~:text=currently%20Advanced%20Moisture%20Diffusion%20Calculation%20Method%20is%20not%20supported%20in%20EnergyPlus.)

honeybee_energy/material/opaque.py
TfedUD: Sounds good! Thanks Chris! should be able to attend to this today/tomorrow

TfedUD: OOOHHHHHH *facepalm*  #still_learning: I got this tho!

PR #825, opened 2022-02-01
feat(properties) add properties that can be extended

PH-Tools:
Following the same pattern from #823 ,  adding a new **.properties** to HB-E [hvac.IdealAirSystem](https://github.com/ladybug-tools/honeybee-energy/blob/4cfb86c040a4587d8bb48715db37749040e82d4c/honeybee_energy/hvac/idealair.py#L15) class in order to allow for extension by plugins.

All hvac tests pass
Note: for consistency, I can add .properties to the rest of the hvac classes if you would like? (allair, doas, heatcool)

### DIFF
--- a/honeybee_energy/hvac/_base.py
+++ b/honeybee_energy/hvac/_base.py
@@ -24,13 +24,12 @@ class _HVACSystem(object):
         * schedules
         * user_data
     """
-    __slots__ = ('_identifier', '_display_name', '_locked', '_user_data', '_properties')
+    __slots__ = ('_identifier', '_display_name', '_locked', '_user_data')
 
     def __init__(self, identifier):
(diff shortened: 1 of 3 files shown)


chriswmackey (changes requested):
Hey @PH-Tools ,

This generally looks good and thanks for adding this. There's just one change that you need to make and then we can merge.

PH-Tools:
Hi @chriswmackey 
Not a problem - that all makes sense. I have updated my commits to reflect those changes - hopefully that all worked properly now?

thanks!

chriswmackey (approved):
LGTM! Let's merge!

PH-Tools:
awesome! Thanks!!

PR #863, opened 2022-04-21
Update .gitignore

PH-Tools:
Would you be averse to adding the following items?
 - *htmlcov* to remove all the coverage html files generated during test runs
 - [*.DS_Store*](https://buildthis.com/ds_store-files-and-why-you-should-know-about-them/) for Mac OS users since no one else would want those

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ tox.ini
 /.vscode
 honeybee.log*
 .env
+htmlcov/*
+.DS_store


mostaphaRoudsari:
Hi @PH-Tools! I merged this in but you can also set a global .gitignore on your side to include what's missing in our repositories.

PH-Tools:
oh! I did not even know a global .gitignore was possible. That is great though: I just added that global to my system for the future.
thanks!

Issue #859, opened 2022-04-15
HBE Properties _load_extension_attr_from_dict:  try...except....  issue


PH-Tools:
Hi

I recently came upon a small issue with the [honeybee_energy.properties.extension._load_extension_attr_from_dict](https://github.com/ladybug-tools/honeybee-energy/blob/96483a27dd838dd3cff59be2a32472a490b55d59/honeybee_energy/properties/extension.py#L52) that I would like to propose a modification for, if it seems correct?

The current code includes:

```python
try:
    setattr(self, '_' + atr, var.__class__.from_dict(property_dict[atr], self.host))
except KeyError:
    pass
```

this works fine to control the flow (ie: if the `property_dict` doesn't have  the `atr`, `from_dict` gets ignored.) but by also wrapping the `from_dict()` function of the actual property itself inside the `try...except...` this catches (and passes) any `KeyErrors` that  occur within the individual property `from_dict()` function.

This is primarily an issue during development, not runtime, but it does mask all KeyErrors made while writing new Properties objects (syntax errors and the like). 

Unless there is an objection, I would suggest replacing the above with something like the following:

```python
atr_prop_dict = property_dict.get(atr, None)
if not atr_prop_dict:
    continue
setattr(self, '_' + atr, var.__class__.from_dict(atr_prop_dict, self.host))
```

unless there is a reason to be passing `KeyErrors` within the actual property `from_dict` functions as well? 
best, 

@PH-Tools

PH-Tools:
Hey @chriswmackey 
You bet - I'll take care of that today. 
best, 
@PH-Tools

PR #867, opened 2022-04-26
Update _load_extension_attr_from_dict func

PH-Tools:
Updating the HBE properties [_load_extension_attr_from_dict](https://github.com/ladybug-tools/honeybee-energy/blob/96483a27dd838dd3cff59be2a32472a490b55d59/honeybee_energy/properties/extension.py#L52) function to remove `try...except: pass` and replace, as per issue #859 

All tests pass, though I did notice that Coverage does not appear to reach these functions inside [properties._EnergyProperties() ](https://github.com/ladybug-tools/honeybee-energy/blob/96483a27dd838dd3cff59be2a32472a490b55d59/honeybee_energy/properties/extension.py#L11)?

If there is a test pattern that you are already using to test the extension functions on HB-Core I'd be happy to replicate that here on HB-Energy? I looked through the HB-Core tests briefly for something like that but didn't find. If not: I can certainly try and come up with something for that for the future as well.

For now, this is tested manually and appears to be working with HBE-Extensions.


chriswmackey (approved):
LGTM, @PH-Tools .

I haven't yet added any tests for these honeybee-energy extension attributes since I added it somewhat quickly to avoid it being a blocker for you. If you wanted to add some, they would be appreciated but you are the main user of these honeybee-energy extension properties right now. So, if you were to find a bug, I trust that you would let us know and/or push a fix.

PR #949, opened 2023-03-21
feat(properties) add .properties to WindowConstructionShade class

PH-Tools:
Add .properties and other required methods to support custom properties

### DIFF
--- a/honeybee_energy/construction/windowshade.py
+++ b/honeybee_energy/construction/windowshade.py
@@ -17,7 +17,7 @@ from ..properties.extension import WindowConstructionShadeProperties
 from honeybee._lockable import lockable
 from honeybee.typing import valid_ep_string
 
-"test"
+
 @lockable
 class WindowConstructionShade(object):
     """Window Construction with shades/blinds or a dynamically-controlled glass pane.
--- a/honeybee_energy/construction/windowshade.py
+++ b/honeybee_energy/construction/windowshade.py
@@ -11,11 +11,13 @@ from ..schedule.dictutil import dict_to_schedule
 from ..schedule.ruleset import ScheduleRuleset
 from ..schedule.fixedinterval import ScheduleFixedInterval
 from ..writer import generate_idf_string
+from ..properties.extension import WindowConstructionShadeProperties
+
 
 from honeybee._lockable import lockable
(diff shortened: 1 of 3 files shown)


PH-Tools:
Hi @chriswmackey 

This PR is to add .properties to WindowConstructionShade, similar to how we did it in PR #861 Hopefully that is ok. I'm trying to get Shades to output to WUFI-Passive correctly. 

Note: all tests pass on my machine except [config_tests :: test_config_ironbug](https://github.com/ladybug-tools/honeybee-energy/blob/c523ebaec8e6ddbfa8af64e6e59dadd66eba76c4/tests/config_test.py#L38)? This test was failing before my edits though. I do not have Ironbug on my Mac, and so this test is not passing?
image



thanks, 
Ed May, @PH-Tools

honeybee_energy/construction/windowshade.py
```diff
@@ -11,11 +11,13 @@
...
 from ..schedule.fixedinterval import ScheduleFixedInterval
 from ..writer import generate_idf_string
+from ..properties.extension import WindowConstructionShadeProperties
+
 
 from honeybee._lockable import lockable
 from honeybee.typing import valid_ep_string
 
-
+"test"
```
chriswmackey, line 20: This looks like an artifact that you forgot to delete.

chriswmackey (commented):
The PR looks good, @PH-Tools , and don't worry about the Ironbug test. As you can see, it passes in the Linux-based CI here. We'll hopefully get Ironbug support on Mac eventually.

I found an artifact in your PR that looks like you forgot to delete. My only other comment might be that you should add a test for your properties here. This will ensure we don't do anything to break these properties with future PRs to the repo.

PH-Tools:
Thanks @chriswmackey , 
- I fixed that "test" typo
- I added some tests for the new WindowConstructionShade properties similar to the [aperture_extend_test](https://github.com/ladybug-tools/honeybee-energy/blob/master/tests/aperture_extend_test.py)

chriswmackey:
LGTM!

chriswmackey (approved):

PH-Tools:
Great! thanks. Do I need to 'close' this PR now on my end to finalize it?

PR #1002, opened 2024-03-12
fix(run): Use new --show-stdout option for OpenStudio CLI

chriswmackey:


honeybee_energy/run.py
```diff
@@ -754,7 +754,7 @@ def _run_osw_unix(osw_json, measures_only=True):
 
     # Write the shell script to call OpenStudio CLI
     measure_str = '-m ' if measures_only else ''
-    shell = '#!/usr/bin/env bash\n"{}" -I "{}" run {}-w "{}"'.format(
+    shell = '#!/usr/bin/env bash\n"{}" -I "{}" run --show-stdout {}-w "{}"'.format(
```
jmarrec, line 757: I'm very curious why you write a shell file then use subprocess to run it, as opposed to directly use subprocess to run the cli command?

jmarrec: (and as a reminder, depending how where this goes, you might want to try --style-stdout and --add-timings)

chriswmackey: Thanks @jmarrec .

Yea, I know that it's not necessarily the most efficient to write out the shell or batch file and then execute it but, if the simulation fails, I've found it useful to just have the file there ready to run the command for debugging. In other words, it's really only there to help me as a developer and I may change this to use the more direct call in the future.

And the formatting you built with the `--style-stdout` is really nice. Super helpful for me understanding where in the process certain warnings come from. I didn't implement it here yet since it seemed like it might be a little too much extra info for our users right now (most of them aren't aware of the OSW we run or the mechanics of OpenStudio CLI). But I will be making use of it and I might expose it to the end user if I can get some of these warnings that might distract them silenced or addressed some other way.

PR #1040, opened 2024-09-21
Add new PV and ElecLoadCenter Properties for Extension

ed-p-may:
Hi @chriswmackey , there are two additional Honeybee-Energy objects that I would like to extend with the usual `.properties`:

- [PvProperties](https://github.com/ladybug-tools/honeybee-energy/blob/8a896d6b4d687118782f3be4a06132a0251c82b0/honeybee_energy/generator/pv.py#L13)
- [ElectricLoadCenter](https://github.com/ladybug-tools/honeybee-energy/blob/8a896d6b4d687118782f3be4a06132a0251c82b0/honeybee_energy/generator/loadcenter.py#L12)

This PR adds the needed `.properties` to both objects, along with the required `to_dict`, `from_dict`, and `duplicate/__copy__` edits. 

Hopefully you agree that this is ok? Note that the naming on the 'PV' one is a little funny: the Honeybee Object itself is already named "PvProperties", even though it is not a normal 'properties' for extension - so I ended up called the extension 'PvPropertiesProperties' which I know is odd, but was trying to follow the pattern from all the others? Not sure what you might think about that though?

Let me know if you think any updates are needed, or you object to this PR for any reason.
Thanks!
@ed-p-may


ed-p-may:
Awesome - thanks @chriswmackey  !

PR #1050, opened 2024-11-15
docs: added idf strings to docstrings of to_idf() functions

TfedUD:
@chriswmackey I triple checked I didn't miss any but could be wrong

### DIFF
--- a/honeybee_energy/simulation/parameter.py
+++ b/honeybee_energy/simulation/parameter.py
@@ -433,6 +433,262 @@ class SimulationParameter(object):
         Args:
             identifier: Text string for to be used as a unique identifier for the
                 IDF Building object.
+                
+            .. code-block:: shell
+
+                !-   ==========================================
+                !-   =========  SIMULATION PARAMETERS =========
+                !-   ==========================================
+
+
+                OutputControl:Table:Style,
+                    CommaAndHTML,             !- column separator
+                    None;                     !- unit conversion
+
+                Output:Variable,
+                    *,                        !- key value
+                    Baseboard Electricity Energy, !- name
+                    Hourly;                   !- frequency
+
+                Output:Variable,
+                    *,                        !- key value
+                    Boiler Electricity Energy, !- name
+                    Hourly;                   !- frequency
+
+                Output:Variable,
+                    *,                        !- key value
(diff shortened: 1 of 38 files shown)


honeybee_energy/simulation/parameter.py
```diff
@@ -433,6 +433,262 @@ def to_idf(self, identifier='Building'):
...
+                Output:Variable,
+                    *,                        !- key value
+                    Zone VRF Air Terminal Heating Electricity Energy, !- name
+                    Hourly;                   !- frequency
+
+                Output:Variable,
+                    *,                        !- key value
+                    Zone Ventilation Fan Electricity Energy, !- name
+                    Hourly;                   !- frequency
+
```
chriswmackey, line 627: Having all of these output vairables here is making the IDF string a little unread-able. What if you just use this code to make the object before you run to_idf:

```
sim_par = SimulationParameter()
sim_par .output.add_output('Zone Ideal Loads Supply Air Total Heating Energy')
sim_par .output.add_output('Zone Ideal Loads Supply Air Total Cooling Energy')
print(sim_par.to_idf()
```

chriswmackey (commented):
Thanks @TfedUD .

This is a big help. I just have one comment about one of the docstrings that is so long that it's a bit unreadable. Once we change that, we should be able to merge.

chriswmackey (changes requested):
Actually, I take it back. It looks like generating the Sphinx docs with all of these new docstrings is giving a bunch of warnings:

image

It looks like a lot of this has to do with improper indentation. I see that this PR also added a bunch of blank lines lines that contain whitespace, which is not following PEP8:

image

You should look at how it is done with the to_dict methods for a sample of how to format these, @TfedUD . Once we can generate the sphinx docs without these warnings and we take out the black whitespaces, we'll be in good shape.

chriswmackey:
Note: the first warning above from Sphinx is my fault. If you fix it, it will be appreciated but just fixing all of the warnings related to to_idf() methods is good enough to merge.

TfedUD:
Hi @chriswmackey sorry about the issues! I'll start working on getting them fixed!

TfedUD:
@chriswmackey  for reference, the last one has the whitespace?
whitespace

Issue #1043, opened 2024-10-13
Add Attributes and Schedules to People


ed-p-may:
Hi @chriswmackey 

I would like to ask it we can add some additional attributes to the [Honeybee-Energy People](https://github.com/ladybug-tools/honeybee-energy/blob/c01d4ebd6186be7507b6f9d34e4df9894538def4/honeybee_energy/load/people.py#L20) object? 

In order to comply with the new [Phius REVIVE standard](https://www.phius.org/phius-revive-2024), they are asking us to set a few specific parameters to these objects for when we do OS simulations. Specifically they are asking for People in the simulations to be able to set:

- Carbon-Dioxide Generation Rate (default=3.82e-08)
- Thermal Comfort Model (default='Pierce')
- Clothing Insulation Schedule
- Air Velocity Schedule
- Work Efficiency Schedule

![Screenshot 2024-10-13 at 2 41 09 PM](https://github.com/user-attachments/assets/57370b41-70f7-4d3e-9083-ffd5de856723)

The intent behind adding these attributes is to enable the output of SET temperature (specifically, the *"Zone Thermal Comfort Pierce Model Standard Effective Temperature"* output variable) from the OSM simulaiton, which they use to verify winter-time thermal resiliency (to certify to their new program, we must show that the total SET degree-hours below 54°F are < 216hrs)

- - -
I'd be happy to put together a PR with this change if you would be ok with this addition? My initial thought is that these do not need to be exposed to the normal user or included in the normal Grasshopper object, and that they can be None in all 'normal' cases - but that they are available for us to add to internally for these new REVIVE calculations.

However If you think there is a better way of enabling these changes without modifying the HBE-People class I can certainly try another method? I originally began by sub-classing the HBE-People class and just adding these new attributes myself - but when the object is serialized/de-serialized as part of the [to-osm workflow](https://github.com/ladybug-tools/honeybee-energy/blob/c01d4ebd6186be7507b6f9d34e4df9894538def4/honeybee_energy/run.py#L331), all that subclass data gets lost since it just de-serialized as a normal HBE-People, so that didn't seem to work.

@ed-p-may

chriswmackey:
Hey @ed-p-may ,

[...]

PR #1146, opened 2026-07-16
feat(people): Add carbon dioxide generation rate

ed-p-may:
## Summary

- add optional `People.carbon_dioxide_generation_rate` in m3/s-W
- preserve existing HBJSON and IDF output byte-for-byte when the value is `None`
- round-trip explicit values through full/abridged dictionaries and the EnergyPlus `People` field 11
- preserve the value through copying, equality/hashing, averaging, and diversification

Related to #1043.

## Backwards compatibility

The property defaults to `None`, which means EnergyPlus uses its default of `3.82e-8 m3/s-W`. When unset, no new dictionary key is emitted and the IDF remains the existing 10-field `People` object. The constructor argument is appended last so existing positional calls remain valid.

## Testing

- `python -m pytest tests/` (`615 passed` on Python 3.13 / OpenStudio 3.10)

The repository CI provisions Ironbug before running the complete suite; the same configuration contract was reproduced locally for the full run.

## Follow-up scope

@chriswmackey, you wrote that you "fully support exposing the Carbon-Dioxide Generation Rate" and offered to handle the corresponding `honeybee-openstudio-gem` work. Since `honeybee-energy` now uses the Python `honeybee-openstudio` package as its active dependency, could you confirm which follow-ups you want and who should own them?

- `honeybee-schema`
- Python `honeybee-openstudio`
- Ruby `honeybee-openstudio-gem`

I have intentionally left schema, translator, and Grasshopper changes out of this PR pending your direction.

### DIFF
--- a/honeybee_energy/load/people.py
+++ b/honeybee_energy/load/people.py
@@ -578,7 +578,7 @@ class People(_LoadBase):
         """A tuple based on the object properties, useful for hashing."""
         return (self.identifier, self.people_per_area, hash(self.occupancy_schedule),
                 hash(self.activity_schedule), self.radiant_fraction,
-                str(self.latent_fraction), str(self.carbon_dioxide_generation_rate))
+                str(self.latent_fraction), self.carbon_dioxide_generation_rate)
 
     def __hash__(self):
         return hash(self.__key())
@@ -592,8 +592,8 @@ class People(_LoadBase):
     def __copy__(self):
         new_obj = People(
             self.identifier, self.people_per_area, self.occupancy_schedule,
-            self.activity_schedule, self.radiant_fraction, self.latent_fraction)
-        new_obj.carbon_dioxide_generation_rate = self.carbon_dioxide_generation_rate
+            self.activity_schedule, self.radiant_fraction, self.latent_fraction,
+            self.carbon_dioxide_generation_rate)
         new_obj._display_name = self._display_name
         new_obj._user_data = None if self._user_data is None else self._user_data.copy()
         new_obj._properties._duplicate_extension_attr(self._properties)
--- a/honeybee_energy/load/people.py
+++ b/honeybee_energy/load/people.py
@@ -48,8 +48,9 @@ class People(_LoadBase):
             estimate the latent fraction based on the occupant's activity level.
             (Default: autocalculate).
         carbon_dioxide_generation_rate: A number greater than or equal to 0 for
-            the carbon dioxide generation rate in m3/s-W. If None, EnergyPlus
-            will use its default of 3.82e-8 m3/s-W. (Default: None).
+            the carbon dioxide generation rate per unit of activity level, in
+            m3/s-W. The default is the EnergyPlus default, which is
+            representative of the average adult. (Default: 3.82e-8).
 
     Properties:
         * identifier
@@ -72,7 +73,7 @@ class People(_LoadBase):
     def __init__(self, identifier, people_per_area,
                  occupancy_schedule=None, activity_schedule=None,
                  radiant_fraction=0.3, latent_fraction=autocalculate,
-                 carbon_dioxide_generation_rate=None):
+                 carbon_dioxide_generation_rate=3.82e-8):
         """Initialize People."""
         _LoadBase.__init__(self, identifier)
         self.people_per_area = people_per_area
@@ -168,8 +169,9 @@ class People(_LoadBase):
 
     @carbon_dioxide_generation_rate.setter
     def carbon_dioxide_generation_rate(self, value):
-        self._carbon_dioxide_generation_rate = None if value is None else \
-            float_positive(value, 'people carbon dioxide generation rate')
+        self._carbon_dioxide_generation_rate = \
+            float_positive(value, 'people carbon dioxide generation rate') \
+            if value is not None else 3.82e-8
 
     @property
     def activity_max_sensible(self):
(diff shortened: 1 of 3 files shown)


honeybee_energy/load/people.py
```diff
@@ -57,23 +60,27 @@ class People(_LoadBase):
...
     __slots__ = ('_people_per_area', '_occupancy_schedule', '_activity_schedule',
-                 '_radiant_fraction', '_latent_fraction')
+                 '_radiant_fraction', '_latent_fraction',
+                 '_carbon_dioxide_generation_rate')
 
     def __init__(self, identifier, people_per_area,
                  occupancy_schedule=None, activity_schedule=None,
-                 radiant_fraction=0.3, latent_fraction=autocalculate):
+                 radiant_fraction=0.3, latent_fraction=autocalculate,
+                 carbon_dioxide_generation_rate=None):
```
chriswmackey: I think you can set this default to `3.82e-8` rather than None and just always write it into the IDF.  It won't have any effect on the simulation to write the default value explicitly and, as the Zen of Python goes, explicit is better than implicit.

honeybee_energy/load/people.py
```diff
@@ -435,6 +458,9 @@ def to_dict(self, abridged=False):
         base['radiant_fraction'] = self.radiant_fraction
         base['latent_fraction'] = self.latent_fraction if \
             isinstance(self.latent_fraction, float) else self.latent_fraction.to_dict()
+        if self.carbon_dioxide_generation_rate is not None:
```
chriswmackey: I like the idea of only writing this key into the dictionary when it is not equal to `3.82e-8`. So just replace this with:
```
if self.carbon_dioxide_generation_rate != 3.82e-8`:
```

honeybee_energy/load/people.py
```diff
@@ -419,6 +439,9 @@ def to_idf(self, zone_identifier):
                     'number of people {ppl}', 'people per floor area {ppl/m2}',
                     'floor area per person {m2/ppl}', 'radiant fraction',
                     'sensible heat fraction', 'activity schedule name')
+        if self.carbon_dioxide_generation_rate is not None:
+            values += (self.carbon_dioxide_generation_rate,)
+            comments += ('carbon dioxide generation rate {m3/s-W}',)
```
chriswmackey: Again, I think it is fine to always write this value explicitly into the IDF.

chriswmackey (changes requested):
Thanks, @ed-p-may .

My main comment is that I don't think we need to support None for this case and we should just always write the number explicitly. This will also make things easier to implement in honeybee-schema given that the honeybee C# bindings derived from that JSON schema struggle with unions between floats and null.

And there is no longer a need to update the `honeybee-openstudio-gem`. That repo is officially deprecated and completely unused in the current software. So the only repos to update are:

* honeybee-openstudio
* honeybee-schema

honeybee_energy/load/people.py
ed-p-may: Done in 086ea13 — the constructor now defaults to `3.82e-8` and the property setter no longer stores `None` (passing `None` just resets to the `3.82e-8` default, following the same pattern as the `Ventilation` setters).

honeybee_energy/load/people.py
ed-p-may: Done in 086ea13 — `to_dict` now uses `if self.carbon_dioxide_generation_rate != 3.82e-8:` exactly as suggested, so existing HBJSON serializations are unchanged. `from_dict`/`from_dict_abridged` fall back to `3.82e-8` when the key is absent.

honeybee_energy/load/people.py
ed-p-may: Done in 086ea13 — the CO2 field is now folded into the main `values`/`comments` tuples so `to_idf` always writes it explicitly. `from_idf` still accepts 10-field legacy strings (or a blank field 11) and falls back to `3.82e-8` for those.

ed-p-may:
Thanks for the review, @chriswmackey! All addressed in 086ea13:

* **No more `None` support.** `carbon_dioxide_generation_rate` now defaults to `3.82e-8` in the constructor, and the setter never stores `None`. The property is always a plain float. This should hopefully make the honeybee-schema follow-up a simple `float` field with a default (no float/null union for the C# bindings to trip over).
* **IDF: always written explicitly.** `to_idf` unconditionally writes field 11. `from_idf` still parses legacy 10-field / blank-field strings and falls back to `3.82e-8`.
* **Dict: key only written when non-default** (`!= 3.82e-8`), as you suggested — existing HBJSON round-trips should match now.
* **Averaging simplified** to a plain weighted sum now that the value is always a number, and the tests + docstring examples were updated to match the new behavior.

Good to know about `honeybee-openstudio-gem` being deprecated. Once this merges, I'm happy to open the follow-up PRs to **honeybee-schema** and **honeybee-openstudio**,  just let me know if you'd rather handle either of those yourself.

Full test suite passing locally (614 passed; the one `config_test.py::test_config_ironbug` failure is a pre-existing local-environment issue unrelated to this change).

chriswmackey (approved):
Thanks, @ed-p-may ,

This version is much cleaner when we don't have to deal with the `None` type. I'll merge this PR once  the other PRs I am merging now have finished making their way through the CI and fee free to get started on PRs to honeybee-openstudio and honeybee-schema.

